PR #3, opened 2023-03-16
feat: chip component #P23-122

mpiwonski:


wlechowicz (approved):

wlechowicz:
I changed `state` prop to `type` since it collides with component state and `type` matches our other conventions better. Quick in-place change for demo purposes. We will discuss later.

PR #4, opened 2023-03-16
feat: side navigation bar item component #P23-124

mpiwonski:
**current challenge**: set up a state for `Icon` component and make it dependable on parent state, so every time side navigation bar item gets clicked, thus goes from inactive to active (vice versa), icon has its color changed.

edit: resolved


packages/ui/Span/index.tsx
Alessiass, line 1: I'm not sure about creating span component. I think that we should use theme from styled-component for such element.
https://reactjsexample.com/a-react-typography-components-with-styled-components/

mpiwonski: Thanks for your insight! I felt like it was not appropriate to create span element, but at that time I had not seen any solution to that. I will definitely look into that, thanks again.

packages/ui/SideNavigationBarItem/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const SideNavigationBarItem = ({
+  children,
+}: SideNavigationBarItemProps) => {
+  const [active, setActive] = React.useState(false);
+
+  return (
+    <SideNavigationBarItemStyled
+      activeFlag={active}
+      onClick={() => {
+        setActive(!active);
```
Noshi96, line 42: When your new state based on the old one, you can use a callback function within access to current previous state value
```javascript 
setActive((prevActiveState) => {return !prevActiveState});
```
[Explanation](https://youtu.be/GGo3MVBFr1A?t=178)

bdeluga: Using a callback in useState is particularly useful when updating state based on asynchronous operations, such as fetching data from an API or responding to user input.
When toggling a boolean state, the new state is always the opposite of the previous state, so using a callback is not strictly necessary.
I would leave it as it is.

mpiwonski: Thanks @Noshi96 and @bdeluga! I appreciate your insights, however I will stick with the current implementation (agreeing more on that with Bartek).

packages/ui/Span/index.tsx
mpiwonski: Going through _styled-typography_ library documentation, I can see there is `Span` component, but what lacks is the text size paramter that I would like to have to be able to control the font size of text inside... Normally, I would just use classes to style the element, but as we are using _styled-components_, I thought I would create own `Span` component and add whatever props I want. Where do we go from here?

wyola (commented):
I think that this should not be a `div` but an `a` (link), since it will be used in the side navigation, and it will lead to different pages (eg. go from "budgets" to "reports"). 

Similar like here, but without the `div` wrapper: https://stackoverflow.com/a/62799750

mpiwonski:
> I think that this should not be a `div` but an `a` (link), since it will be used in the side navigation, and it will lead to different pages (eg. go from "budgets" to "reports").
> 
> Similar like here, but without the `div` wrapper: https://stackoverflow.com/a/62799750

Thanks for your input, Ola :) That's what I'll be going for, but maybe it could be Link component from next/link?

AzulloCyll:
I thintk that actice element indicator can't be just border-left css property. Look closer to it.
Untitled-1

mpiwonski:
Thanks Daniel for your comment! That's right. What I'm going to do probably
is wrap icon and span elements in one div (these will be flex elements with
flex direction column) and make another div for this border like element -
so these two divs will also be flex, but in a row manner.

czw., 23 mar 2023, 08:21 użytkownik Daniel Chmur ***@***.***>
napisał:

packages/ui/SideNavigationBarItem/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+          textValue={textValue}
+          icon={icon}
+        >
+          <Icon
+            icon={icon}
+            iconSize={30}
+            color={activeFlag ? "#1e4c40" : "#7e7e7e"}
+          />
+          <SpanStyled fontSize={10}>{textValue}</SpanStyled>
+          {activeFlag && <DivStyled />}
```
bdeluga, line 27: That's a cleaver way of doing that border. Nice 💪 .

packages/ui/SideNavigationBarItem/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+        </SideNavigationBarItemStyled>
+      </LinkStyled>
+    </ListItemStyled>
+  );
+};
+
+export type SideNavigationBarItemProps = {
```
aniasgit, line 34: Move the props type above the component, it's easier to read.

aniasgit (approved):
Looks good to me :)

packages/ui/SideNavigationBarItem/index.tsx
Noshi96, line 1: As @aniasgit mentioned, when it comes to code readability, it's a good idea to keep some structure, e.g. declare all types/interfaces first, then the component itself and under it the styles. And if you declare styledcomponentns, you can also keep the order there. In your case, ListItemStyled should be first, then LinkStyled, and then SideNavigationBarItemStyled.

packages/ui/SideNavigationBarItem/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+}: SideNavigationBarItemProps) => {
+  return (
+    <ListItemStyled>
+      <LinkStyled href={href}>
+        <SideNavigationBarItemStyled
+          activeFlag={activeFlag}
+          href={href}
+          textValue={textValue}
+          icon={icon}
+        >
```
Noshi96, line 20: In your SideNavigationBarItemStyled you only use activeFlag, I wonder if we need to pass all these values to this single div. I think you can add an addidtional interface instead:
```typescript
interface SideNavigationBarItemState {
  activeFlag: boolean;
}
``` 

Either way, you keep access to all destroyed values in nested divs

packages/ui/SideNavigationBarItem/index.tsx
mpiwonski: Good to be resolved 👍

packages/ui/SideNavigationBarItem/index.tsx
mpiwonski: Fixed.

packages/ui/SideNavigationBarItem/index.tsx
mpiwonski: Thanks, good point! Fixed.

Noshi96 (approved):

packages/ui/SideNavigationBarItem/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  href: string;
+  textValue: string;
+} & React.HTMLProps<HTMLDivElement> &
+  IconProps;
+
+type SpanProps = {
+  fontSize: number;
+} & React.HTMLProps<HTMLSpanElement>;
+
+interface SideNavigationBarItemState {
```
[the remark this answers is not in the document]
mpiwonski: Yeah, @Noshi96 has suggested to use `interface` here https://github.com/intive/patronage2023-js/pull/4#discussion_r1149280231. It's actually a good question if we could just simply use another `type` :)

apps/storybook/stories/SideNavigationBar.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+  items: [
+    { href: "#menu", icon: "menu", textValue: "Menu" },
+    { href: "#history", icon: "history", textValue: "History" },
+    { href: "#settings", icon: "settings", textValue: "Settings" },
+    {
+      href: "#shopping_cart",
+      icon: "shopping_cart",
+      textValue: "Shopping cart",
+    },
+  ],
```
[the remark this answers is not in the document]
mpiwonski: True, but actually I have not done it by myself, but Prettier has to take credit for that 🐙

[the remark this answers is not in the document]
zuzannapaw: Oh okay! ☺️

packages/ui/SideNavigationBarItem/index.tsx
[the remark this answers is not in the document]
zuzannapaw: Thanks! I can see it now ☺️

packages/ui/SideNavigationBar/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  pathname: string;
+};
+
+export const SideNavigationBar = ({
+  items,
+  pathname,
+}: SideNavigationBarProps) => {
+  return (
+    <SideNavigationBarStyled items={items} pathname={pathname}>
+      {items.map(({ href, icon, textValue }, index) => (
```
zuzannapaw, line 18: I have to tell, I really liked how you used object destructuring here! Well done 💪

zuzannapaw (approved):

PR #7, opened 2023-03-16
feat: card component #P23-114

aniasgit:


apps/storybook/stories/Card.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
```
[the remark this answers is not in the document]
aniasgit: You're right, I already changed it

packages/ui/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
+"use client";
+
+import { ReactNode } from "react";
+import styled, { css } from "styled-components";
```
wyola, line 4: { css } is not used - you might think about removing it

packages/ui/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+	verticalFit,
+}: CardProps) => {
+	return (
+		<CardWrapperStyled horizontalFit={horizontalFit} verticalFit={verticalFit}>
+			<CardStyled
+				padding={padding}
+				margin={margin}
+				horizontalFit={horizontalFit}
+				verticalFit={verticalFit}>
+				<CardContenetStyled>{children}</CardContenetStyled>
```
wyola, line 28: typo - "Contenet"

wyola (commented):
I think that this component could be just a single `div` with padding, so that content does not touch the borders. Padding changes on different screen sizes, so there could be a media query in the styling.

bdeluga (commented):
unused css from styled-components and small typo, fix quickly and we can approve 🚀

packages/ui/Card/index.tsx
aniasgit: done

packages/ui/Card/index.tsx
aniasgit: In the latest changes, I removed this part.

AzulloCyll:
In figma there is 1px border on Card component when it's on white background but there is no border when Card is on dark. 
I suggest adding border prop. Maybe with default value.

AzulloCyll (approved):
Looks good form me

zuzannapaw (approved):

PR #12, opened 2023-03-18
feat: input component #P23-117

wyola:


packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+        placeholder=" "
+        hasError={hasError}
+        id={id || randomId}
+        type={typeOverride || type}
+        value={value}
+        onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+          if (onChange) {
+            onChange(event.target.value);
+          }
+        }}
```
bdeluga, line 91:
```suggestion
onChange={onChange}
```
Someone that uses this input probably knows how he need to handle onChange. Let's just pass their function instead.🚀

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+            inputRef.current?.focus();
+          }}
+        >
+          <Icon
+            icon="visibility"
+            color="#49454f"
+            iconSize={23}
+          />
+        </StyledIcon>
+      );
```
bdeluga, line 54:
```suggestion
return (
<StyledIcon
         onClick={() => {
            setPasswordShown(!passwordShown)
            inputRef.current?.focus()
          }}
        >
          <Icon
            icon={`${passwordShown ? 'visibility_off' : 'visibility'}`}
            color="#49454f"
            iconSize={23}
          />
</StyledIcon>
)
```
This code works great, but i think you overdid it. What would you say we cut this code in half ? ✂️ 
Of course we have to remember to change how we handle Input type with this change

bdeluga (changes requested):

packages/ui/Input/index.tsx
wyola: Thanks, I did this change, just slightly different on one line, since typescript was mad 😅

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+  hasError: boolean;
+} & React.HTMLProps<HTMLInputElement>;
+
+type StyledLabelProps = {
+  hasError: boolean;
+} & React.HTMLProps<HTMLLabelElement>;
+
+type StyledSupportingLabelProps = {
+  hasError: boolean;
+} & React.HTMLProps<HTMLDivElement>;
```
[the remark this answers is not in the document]
wyola: thanks for that, I changed the code

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,190 @@
+import * as React from "react";
```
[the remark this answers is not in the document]
wyola: I was importing it like this, since I was using it as well with things like `React.HTMLProps<HTMLInputElement>` etc. but after changing those parts of code according to your other remarks, I can also change this one 🫡

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+          }
+        }}
+      />
+      <StyledLabel hasError={hasError} htmlFor={id || randomId}>
+        {label}
+      </StyledLabel>
+      {getButton()}
+      <StyledSupportingLabel hasError={hasError}>
+        {supportingLabel}
+      </StyledSupportingLabel>
```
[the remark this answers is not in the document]
wyola: I changed it, but this `<StyledSupportingLabel hasError>` has to be like this `<StyledSupportingLabel hasError={hasError}>` because otherwise it would always be true and display red supporting label :) 
image

but thank you for pointing this out in general, because I got so focused on this supporting label that indeed I forgot that there are inputs without it :)

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+        </StyledIcon>
+      );
+    }
+    if (value) {
+      return (
+        <StyledIcon
+          onClick={() => {
+            if (onChange) {
+              onChange("");
+            }
```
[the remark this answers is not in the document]
wyola: it accepts only strings, which is defined in props `onChange?: (newValue: string) => void;`

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+  label,
+  hasError = false,
+  id,
+  supportingLabel,
+  type,
+  value,
+  onChange,
+}: InputProps) => {
+  const randomId = useId();
+  const [typeOverride, setTypeOverride] = useState("");
```
Noshi96, line 16: If your useState only takes two values ("" and "text"), can it be a boolean type?

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+      )}
+    </Wrapper>
+  );
+};
+
+type StyledErrorProps = {
+  hasError: boolean;
+};
+
+export type InputProps = {
```
monika-klokosinska-intive, line 93: @wyola You use `InputProps` in line 14, but define it here. The code would be more readable if `InputProps` were defined before its first usage.

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+      <StyledInput
+        ref={inputRef}
+        placeholder=" "
+        hasError={hasError}
+        id={id || randomId}
+        type={typeOverride || type}
+        value={value}
+        onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+          if (onChange) {
+            onChange(event.target.value);
```
monika-klokosinska-intive, line 72: @wyola I would pass the whole event to `onChange`. You don't know what the passed function needs from the the event and it's not really the concern of this component. The function can then extract what it needs from the event and use it, but that's outside the scope of responsibility for the input.

monika-klokosinska-intive: @wyola 
```
 if (onChange) {
            onChange(event);
 ```

File: packages/ui/Input/index.tsx [hunk 10]
```diff
@@ -0,0 +1,168 @@
...
+      background-color: white;
+      padding-left: 4px;
+      padding-right: 4px;
+    }
+  }
+
+  ${({ hasError }) =>
+    hasError &&
+    css`
+      border-color: #b3261e !important;
```
monika-klokosinska-intive, line 133: `!important` is very bad, not good, terrible practice. What is the reason for it here?

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+  );
+};
+
+type StyledErrorProps = {
+  hasError: boolean;
+};
+
+export type InputProps = {
+  label: string;
+  onChange?: (newValue: string) => void;
```
monika-klokosinska-intive, line 95: @wyola how about `onChange?: React.ChangeEventHandler<HTMLInputElement>`. This way it's still typed, but still allows for more freedom in how the function will be built (see my other comments concerning `onChange`)

packages/ui/Input/index.tsx
wyola: @Noshi96  it could be, it's a possibility, but in my opinion, it does not change much

packages/ui/Input/index.tsx
monika-klokosinska-intive: @wyola `!important` should always be the last resort after everything else failed. Otherwise it gets over-used and before you know, half of your css rules have `!important` in them. Overwriting them then (if you need to) becomes impossible without sticking more `!important`... and you can probably see where it is going.
It's a useful tool in case some extreme cases, but shouldn't be the first thing you reach for.

wyola: @monika-klokosinska-intive ok, I fixed it, will push it soon with other changes

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+  hasError: boolean;
+};
+
+export type InputProps = {
+  label: string;
+  supportingLabel?: React.ReactNode;
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  onInputCleared?: () => void;
+} & StyledErrorProps &
+  React.HTMLProps<HTMLInputElement>;
```
wyola, line 16: @monika-klokosinska-intive @wlechowicz should I limit the props from input element only to those that are used by the component? https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

it could look like this:
```suggestion
  Pick<React.HTMLProps<HTMLInputElement>, 'id' | 'type' | 'value'>;
```

monika-klokosinska-intive: @wyola either way is fine with me! @wlechowicz what do you think?

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+    if (type === "password") {
+      return (
+        <StyledIcon
+          onClick={() => {
+            setTypeOverride(typeOverride ? "" : "text");
+            inputRef.current?.focus();
+          }}
+        >
+          <Icon
+            icon={`${typeOverride ? "visibility_off" : "visibility"}`}
```
monika-klokosinska-intive, line 49: @wyola you can achieve the same thing with simpler code e.g ` icon={typeOverride ? "visibility_off" : "visibility"`.

monika-klokosinska-intive:
@wyola Great job on all the changes in PR! 
Looks like you have a conflict in `packages/ui/index.tsx` - please merge the main branch to yours and resolve it.

wyola:
> @wyola Great job on all the changes in PR! Looks like you have a conflict in `packages/ui/index.tsx` - please merge the main branch to yours and resolve it.

@monika-klokosinska-intive thank you! I merged it and resolved the conflict. One of the checks fails and I can't seem to find why (it was fine before)

image

AzulloCyll (approved):
I think it's good to merge.

packages/ui/Input/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+    if (value && onInputCleared) {
+      return (
+        <StyledIcon
+          onClick={() => {
+            onInputCleared();
+            inputRef.current?.focus();
+          }}
+          style={{ transform: "rotate(45deg)" }}
+        >
+          <Icon icon="add_circle" color="#49454f" iconSize={20} />
```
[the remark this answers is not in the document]
wyola: @aniasgit you are right of course, I was trying to find this icon with "close" and "x", did not think to type "cancel" 😂 🤦🏻‍♀️will change it now

wyola:
> > @wyola Great job on all the changes in PR! Looks like you have a conflict in `packages/ui/index.tsx` - please merge the main branch to yours and resolve it.
> 
> @monika-klokosinska-intive thank you! I merged it and resolved the conflict. One of the checks fails and I can't seem to find why (it was fine before)
> 
> image

nvm, I fixed it (it was missing "use client" and web build was failing)

packages/ui/Input/index.tsx
Noshi96, line 1: Input component does not match the design (padding and green icon)
image

wyola: @Noshi96 what you are showing there is distance to line height, but nevertheless you are right and padding should be 20px top and bottom and 16px left 
image

wyola: @Noshi96 I corrected it, changed CSS a bit to include line height to be 150% of a font size and padding to be 14px (since 16px is together with the 2px border in the design).

Good catch btw!

PawelPazderski:
Can we pls add (optional) onFocus to the input? I'll need it in Log In form to close the error message for sure.

bdeluga:
> Can we pls add (optional) onFocus to the input? I'll need it in Log In form to close the error message for sure.

It alread has optional onFocus since it extends normal input right ?
image

bdeluga:
Like i did above ?

PawelPazderski:
> Like i did above ?

Yep, got this:
image

bdeluga:
Hmm, did you added "use client"

PawelPazderski:
> Hmm, did you added "use client"

Yes, go use client I test it in the "Buttons"
image

PawelPazderski:
> So is it working, cause I got confused right now haha 😅

It's not on my side, if the onFocus is not added to props

bdeluga:
I downloaded commit with onFocus that Ola already added. (no wonder it was working 😅 ) Silly me forgot we aren't passing all props `{...props}` sorry for confuse. @PawelPazderski  is 100% right onFocus is needed.

PawelPazderski (approved):
Looks good works fine to me now :)

PR #13, opened 2023-03-19
feat: link component #P23-112

Vectari:


packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+};
+
+export const LinkComponentAnchorStyled = styled.a<LinkComponentProps>`
+  /* position: absolute; */
+  width: 57px;
+  height: 24px;
+  left: 0px;
+  top: 0px;
```
aniasgit, line 35: With the default static position, the top and left settings do not work, similarly with a inline element, the width and height do not work.

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  line-height: 24px;
+  /* identical to box height, or 150% */
+  color: #397B65;
+  text-decoration: underline;
+`;
+
+export const LinkComponentButtonStyled = styled.button<LinkComponentProps>`
+  /* position: absolute; */
+  left: 0px;
+  top: 0px;
```
aniasgit, line 49: https://github.com/intive/patronage2023-js/pull/13/files#r1141929818

packages/ui/index.tsx
```diff
@@ -1,2 +1,3 @@
 import * as React from "react";
 export { Button } from "./Button";
+export { LinkComponent } from "./Link";
```
monika-klokosinska-intive, line 3: @aniasgit Github krzyczy, że każdy plik powienien się kończyć nową linią. Jeśli się nie kończy, to niektóre procesy mogą zgłupieć. U nas się to raczej nie wydarzy, ale i tak dobrą praktyką jest wstawiać tę nową linię na końcu pliku. 
Jeśli używasz Visual Studio Code, to w Settings jest opcja `Files: Insert Final New Line`. Po zaznaczeniu VSC sam będzie dodawał te linię przy zapisywaniu pliku, github będzie zadowolony, a TY nie będziesz musieć o tym pamiętać.

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  href?: any;
+  label?: string;
+} & React.HTMLProps<HTMLElement>;
+
+export const LinkComponent = ({ 
+  href, 
+  children,
+  onClick,
+}: LinkComponentProps) => {
+  if (onClick) {
```
monika-klokosinska-intive, line 14: You could use a trenary operator to make the code shorter:
```
return onClick ? (<LinkComponentButtonStyled onClick={onClick}>
        {children}
      </LinkComponentButtonStyled>) :     (<LinkComponentAnchorStyled href={href}>
        {children}
      </LinkComponentAnchorStyled>)
```
I leave the final decision to you - please choose the version that makes the component more readable in your option :).

packages/ui/index.tsx
Vectari: Done :)

packages/ui/Link/index.tsx
Vectari: Less code is always better. I changed to ternary operator. Thanks for the advice!

packages/ui/Link/index.tsx
Vectari: I added position: absolute and now top and left are working. In my tests width and height for anchor and button are not important but they are defined in design (
with a few inaccuracies :) ) so i leave then in comment because thay may be useful later.

packages/ui/Link/index.tsx
Vectari: I added position: absolute and now top and left are working. In my tests width and height for anchor and button are not important but they are defined in design (
with a few inaccuracies :) ) so i leave then in comment because thay may be useful later.

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  font-style: normal;
+  font-weight: 400;
+  font-size: 16px;
+  /* identical to box height, or 150% */
+  color: #397B65;
+  width: 54px;
+  height: 20px; 
+  line-height: 20px;
+  flex: 1;
+  flex-grow: 0;
```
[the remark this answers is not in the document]
Vectari: I added display: flex for button. Thanks for the review!

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,55 @@
+"use client";
+import styled from "styled-components";
+
+type LinkComponentProps = {
+  href?: any;
+  label?: string;
```
aniasgit, line 6: I don't see any place where you use label, do we need it?

Vectari: I was using label before children and forget to change above. Weird that I don’t had errors. I fix this after work :)

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+    {children}
+  </LinkComponentButtonStyled>) : (<LinkComponentAnchorStyled href={href}>
+    {children}
+  </LinkComponentAnchorStyled>)
+};
+
+export const LinkComponentAnchorStyled = styled.a<LinkComponentProps>`
+  position: absolute;
+  top: 0px;
+  left: 0px;
```
monika-klokosinska-intive, line 24: @Vectari a link doesn't need `position: absolute` or `top` and `left` values - it's position should be decided by parent component. Please remove them!

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  /* width: 57px; */
+  /* height: 24px; */
```
monika-klokosinska-intive, line 26: @Vectari it's bad practice to leave unused code in comments. Better just to remove it altogether, to avoid lengthy messy code later on.

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  font-size: 16px;
+  line-height: 24px;
+  color: #397B65;
+  text-decoration: underline;
+`;
+
+export const LinkComponentButtonStyled = styled.button<LinkComponentProps>`
+  position: absolute;
+  top: 0px;
+  left: 0px;
```
monika-klokosinska-intive, line 39: @Vectari similar to Link, the button shouldn't need `position: absolute;` or `top` and `left`. The parent using it will be responsible for putting it in the right place.

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  /* width: 54px; */
+  /* height: 20px;  */
```
monika-klokosinska-intive, line 41: @Vectari here as well, if you don't need a piece of value altogether, better remove it altogether!

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  font-family: 'Inter';
+  font-style: normal;
+  font-weight: 400;
+  font-size: 16px;
+  line-height: 20px;
+  color: #397B65;
+  display: flex;
```
monika-klokosinska-intive, line 48: All flex styles for the button are unnecessary here - it will look the same without them. I assume you picked these styles from Figma - you need to be careful with things like position, flex, grid. The figma css are generated automatically and the styles for placing elements in other elements can sometimes be either trash, unnecessary or overkill (e.g. figma generated a complicated css grid for something that flex could handle). With things like fonts, colors, borders you can usually copy 1:1, but with others you need to ask yourself it it's really the best way to style them.

monika-klokosinska-intive:
@Vectari You have got a conflict! Please merge main branch to yours and resolve it :)!

packages/ui/Link/index.tsx
Vectari: I fixed it :)

packages/ui/Link/index.tsx
Vectari: I fixed it :)

packages/ui/Link/index.tsx
Vectari: I fixed it :)

packages/ui/Link/index.tsx
Vectari: I fixed it :)

packages/ui/Link/index.tsx
Vectari: I fixed it :)

monika-klokosinska-intive (approved):

packages/ui/Link/index.tsx
```diff
@@ -0,0 +1,38 @@
+"use client";
```
Noshi96, line 1: In figma we have another link, "forgot password?" it is similar to the button link, because there is no underline, but I wonder if we need such a variant, because there is no design for next steps, for example, "password recovery screen".

Vectari: You are right but without design is hard to tell how to resolve this. Most of the websites lead to new page for recovery.
I have to ask leaders.

aniasgit (approved):

PR #17, opened 2023-03-23
feat: avatar component #P23-111

wyola:


packages/ui/Avatar/index.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from "styled-components";
+
+export type AvatarProps = {
+  className?: string;
+  username: string;
+  src: string;
+};
+
+export const Avatar = ({ username, src, className }: AvatarProps) => {
```
wlechowicz, line 9:
```suggestion
export const Avatar = ({ username = 'user', src, className }: AvatarProps) => {
```

apps/storybook/stories/Avatar.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+const Wrapper = styled.div`
+  height: 48px;
+  display: flex;
+  align-items: center;
+  font-size: 14px;
+  font-weight: 400;
+`;
+
+const StyledAvatar = styled(Avatar)`
+  margin-right: 16px;
```
wlechowicz, line 51: Positioning of the Avatar elements (grid in user profile, collapsed in table) is the responsibility of the parent, not the Avatar itself. So this spacing between Avatar and name should be done with the parent's flex props, like gap.

apps/storybook/stories/Avatar.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const Template = ({ height, ...args }: StoryProps) => (
+  <div style={{ height: `${height}px` }}>
+    <Avatar {...args} />
+  </div>
+);
+
+export const Normal = Template.bind({});
+Normal.args = {
+  height: 120,
```
wlechowicz, line 22: This needs to be disabled as a control in the story, because it will show up in storybook as a prop of the Avatar, but it's not - it will confuse others.

wlechowicz (changes requested):

packages/ui/Avatar/index.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from "styled-components";
+
+export type AvatarProps = {
+  className?: string;
+  username: string;
+  src: string;
+};
```
Noshi96, line 7: You can add ```& React.HTMLProps<HTMLImageElement>``` and skip the type declaration for ```src``` and ```className```

packages/ui/Avatar/index.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from "styled-components";
+
+export type AvatarProps = {
+  className?: string;
+  username: string;
+  src: string;
```
monika-klokosinska-intive, line 6: @wyola I believe we also discussed adding `alt` as a prop (for the screen with all avatars where the user selects one). The logic on the image would be along the lines of `alt={alt ? alt : ${username}'s avatar}` then.

wyola: @monika-klokosinska-intive done

packages/ui/Avatar/index.tsx
wyola: @wlechowicz done

apps/storybook/stories/Avatar.stories.tsx
wyola: @wlechowicz it was removed at all after changes we discussed on slack

aniasgit (approved):
Looks good to me 👍

zuzannapaw (approved):

apps/storybook/stories/Avatar.stories.tsx
```diff
@@ -9,44 +9,42 @@ export default {
...
-  font-weight: 400;
-`;
-
-const StyledAvatar = styled(Avatar)`
-  margin-right: 16px;
-`;
+  font-size: 48px;
+  line-height: 1em;
+  
+  > * + * {
```
mpiwonski, line 47: Nicely done here! 👍

mpiwonski (approved):

packages/ui/Avatar/index.tsx
wlechowicz: Contrary to things like button or link, we don't want the Avatar to accept all the properties of img, such as width and height, and we have no intention of passing them down to the img element. Therefore it's fine to declare our own prop types without extending image element.

wlechowicz (changes requested):
Looks good overall, one minor change requested.

apps/storybook/stories/Avatar.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+const Wrapper = styled.div`
+  display: flex;
+  align-items: center;
+  font-size: 48px;
+  line-height: 1em;
+  
+  > * + * {
+    margin-left: -0.4em;
+  }
+`;
```
monika-klokosinska-intive, line 50: @wyola new line please!

packages/ui/Avatar/index.tsx
```diff
@@ -0,0 +1,39 @@
+import styled, { css } from "styled-components";
+
+export type AvatarProps = {
+  username?: string;
+  src: string;
+  alt?: string;
+  hasOutline?: boolean;
```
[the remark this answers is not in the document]
wyola: @wlechowicz done

apps/storybook/stories/Avatar.stories.tsx
wyola: @monika-klokosinska-intive done

monika-klokosinska-intive (approved):

wlechowicz (approved):

PR #18, opened 2023-03-26
feat: sign-in page layout #P23-150

AzulloCyll:
Added simple global css reset file to get rid of unneccesary margin on <body> and set box-sizing: border-box; on all elements.

Go to `http://localhost:3000/sign-in` or `https://patronage2023-js-e70s01pvt-wlechowicz.vercel.app/sign-in` to see result of my work.


apps/web/app/layout.tsx
```diff
@@ -21,10 +21,14 @@ export default function RootLayout({ children }: LayoutProps) {
...
         />
+        <link
+          rel="stylesheet"
+          href="https://fonts.googleapis.com/css2?family=Signika:wght@600&display=swap"
+        />
       </head>
       <body className={inter.className}>
         <StyledComponentsRegistry>
-          <Nav />
+          {/* <Nav /> */}
```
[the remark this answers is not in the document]
Alessiass: The change is good still. The Nav should be completely removed from root layout since it only occurs in certain pages.

apps/web/app/sign-in/layout.tsx
Alessiass, line 1: The layout location should be changed since it's not only used in sign-in page.

Alessiass (changes requested):

apps/web/app/sign-in/layout.tsx
```diff
@@ -0,0 +1,112 @@
...
+export default function SignInLayout({ children }: LayoutProps) {
+  return (
+    <ViewStyled>
+      <Background>
+        <PageContainerStyled>
+          <ContentStyled>
+            <SectionStyled>
+              <Logo logoWidth={138} white />
+              <StyledH1>Log in with your email</StyledH1>
+              <StyledP>Use your email to log in to your team workspace</StyledP>
```
zuzannapaw, line 24: What do you think of instead of `<StyledP>` use full name of tags like `<StyledParagraph>`? It could help with clarity of component's names. 😊

apps/web/app/sign-in/layout.tsx
```diff
@@ -0,0 +1,112 @@
...
+  padding: 8px;
+
+  @media ${device.desktop} {
+    height: 100vh;
+    padding: 12vh 64px;
+  }
+`;
+
+const StyledH1 = styled.h1`
+  display: block;
```
zuzannapaw, line 89: I think `h1` has display block by it's nature 🔲🌳 , so you don't need to worry about adding it in styles.

apps/web/app/layout.tsx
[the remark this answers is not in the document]
AzulloCyll: So I've removed Nav completly.

apps/web/app/sign-in/page.tsx
```diff
@@ -0,0 +1,14 @@
...
+        Zawieszony nad jak Śmiga w otwartą dłoń, a on pod gruzami... zarazi W
+        słabo połyskujących szkłach czytelne znaki obcego Oczywiście był to
+        flankowały rzędy foteli. Trwał między obie skarpetki, na wystawie gazowe
+        pistolety, się Instytut i szło Łączka? Wykładowca założył że wkrótce
+        pojawią się Najbliższą pokrywały czarne zakręcał, gdy atlasik był
+        dozwolony owe brzmiące jak kombinezonie na stalowej desce pytanie Kadet
+      </p>
+    </div>
+  );
+}
```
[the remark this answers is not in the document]
AzulloCyll: Ok. Changed to Ipsum. :D

apps/web/app/sign-in/layout.tsx
AzulloCyll: Thanks. I've changed it to ParagraphStyled due to other elements naming convention.

apps/web/app/sign-in/layout.tsx
AzulloCyll: Thats true. Deleted this line.

apps/web/app/sign-in/layout.tsx
AzulloCyll: Moved my layout to auth folder, it can serve sign-in and sign-up.
I'm still thinking about that solution.

apps/web/app/auth/sign-in/layout.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { Card, Logo } from "ui";
+import { LayoutProps } from "../layout";
+import {device} from "lib/css-variables"
+
+export default function SignInLayout({ children }: LayoutProps) {
+  return (
+    <ContentStyled>
+      <SectionStyled>
+        <Logo logoWidth={138} white />
+        <StyledH1>Log in with your email</StyledH1>
```
aniasgit, line 13: I think that because of naming convention H1Styled would be better.

apps/web/app/auth/sign-in/page.tsx
```diff
@@ -0,0 +1,20 @@
+export default function SignInPage() {
+  return (
+    <div>
+      <p style={{ margin: "50px" }}>
```
aniasgit, line 4: Instead of margin here, you can use padding in Card component

apps/web/app/auth/sign-in/layout.tsx
AzulloCyll: Right. Thanks

apps/web/app/auth/sign-in/layout.tsx
```diff
@@ -0,0 +1,85 @@
...
+    padding-left: 128px;
+    padding-right: 64px;
+  }
+`;
+
+const FormWrapperStyled = styled.div`
+  width: 100%;
+  height: 100%;
+  padding: 8px;
+
```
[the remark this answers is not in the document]
AzulloCyll: Thanks, it works good.

apps/web/lib/css-variables.tsx
```diff
@@ -0,0 +1,4 @@
+export const device = {
+  mobile: `(max-width: 767px)`,
+  desktop: `(min-width: 768px)`,
+};
```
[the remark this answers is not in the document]
AzulloCyll: added new line

monika-klokosinska-intive (approved):

wlechowicz (approved):

wlechowicz (changes requested):
Actually, one change to be made.
The layout that you put in `app/(regflow)/sign-in/layout.tsx` will be the common layout for both sign in and sign up, so it should all be the de facto regflow layout located in `app/(regflow)/layout.tsx` as every page under (regflow) group will use this layout, not just the sign-in page.

aniasgit (approved):
Looks good 👍

wlechowicz (approved):
This will be refactored as a follow up - layout needs to be made server side, so all the styled components need to be moved out of the layout iself.
We will also explore better solutions rather than usePathname for the left column content.

PR #19, opened 2023-03-27
hotfix: buttongroup-render method #P23 121  

bdeluga:
Had to use `import {Fragment} from 'react'` to apply key, because apparently browser was smoking red, when using my component.
Also used `Icon` in story instead of `<span>` for icon.

### DIFF
--- a/apps/storybook/stories/ButtonGroup.stories.tsx
+++ b/apps/storybook/stories/ButtonGroup.stories.tsx
@@ -87,6 +87,7 @@ Secondary.args = {
       onSelect: () => {
         console.log("Option 2 selected");
       },
+      defaultChecked: true,
     },
   ],
   secondary: true,
--- a/apps/storybook/stories/ButtonGroup.stories.tsx
+++ b/apps/storybook/stories/ButtonGroup.stories.tsx
@@ -70,7 +70,7 @@ Primary.args = {
       onSelect: () => {
         console.log("Option 3 selected");
       },
-      id: "expanses",
+      id: "expenses",
     },
   ],
   secondary: false,
@@ -92,7 +92,7 @@ Secondary.args = {
         console.log("Option 2 selected");
       },
       defaultChecked: true,
-      id: "expanses",
+      id: "expenses",
     },
   ],
   secondary: true,
--- a/apps/storybook/stories/ButtonGroup.stories.tsx
+++ b/apps/storybook/stories/ButtonGroup.stories.tsx
@@ -1,7 +1,7 @@
 import { ComponentStory, ComponentMeta } from "@storybook/react";
 import React from "react";
 
-import { ButtonGroup } from "ui";
+import { ButtonGroup, Icon } from "ui";
 
 export default {
   title: "ButtonGroup",
@@ -9,7 +9,7 @@ export default {
   argTypes: {},
 } as ComponentMeta<typeof ButtonGroup>;
 
-const FullWidth: ComponentStory<typeof ButtonGroup> = (args) => (
+const FullWidth: ComponentStory<typeof ButtonGroup> = args => (
   <div>
     <p>
       This component <u>always</u> fills parent. Both height and width.
@@ -18,7 +18,7 @@ const FullWidth: ComponentStory<typeof ButtonGroup> = (args) => (
   </div>
 );
 
-const FixedWidth: ComponentStory<typeof ButtonGroup> = (args) => (
+const FixedWidth: ComponentStory<typeof ButtonGroup> = args => (
   <>
     <p>Parent has fixed width (300px) and component is secondary.</p>
     <div
@@ -31,7 +31,7 @@ const FixedWidth: ComponentStory<typeof ButtonGroup> = (args) => (
   </>
 );
 
-const FullHeight: ComponentStory<typeof ButtonGroup> = (args) => (
+const FullHeight: ComponentStory<typeof ButtonGroup> = args => (
   <>
     <p>
       It also takes icons from material-symbol library. In this example parent
@@ -97,13 +97,13 @@ export const Icons = FullHeight.bind({});
 Icons.args = {
   options: [
     {
-      component: <span className="material-symbols-rounded">equalizer</span>,
+      component: <Icon icon="bar_chart" />,
       onSelect: () => {
         console.log("Option 1 selected");
       },
     },
     {
-      component: <span className="material-symbols-rounded">area_chart</span>,
+      component: <Icon icon="area_chart" />,
       onSelect: () => {
         console.log("Option 2 selected");
       },
--- a/apps/storybook/stories/ButtonGroup.stories.tsx
+++ b/apps/storybook/stories/ButtonGroup.stories.tsx
@@ -9,7 +9,7 @@ export default {
   argTypes: {},
 } as ComponentMeta<typeof ButtonGroup>;
 
-const FullWidth: ComponentStory<typeof ButtonGroup> = args => (
+const FullWidth: ComponentStory<typeof ButtonGroup> = (args) => (
   <div>
     <p>
       This component <u>always</u> fills parent. Both height and width.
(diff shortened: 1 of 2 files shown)


mpiwonski (changes requested):

packages/ui/ButtonGroup/index.tsx
```diff
@@ -1,41 +1,41 @@
...
               onClick={onSelect}
-              name={'button-group'}
+              name={"button-group"}
             />
             <label key={`index-${index}`} htmlFor={`button-${index}`}>
               {component}
             </label>
-          </>
-        )
+          </Fragment>
```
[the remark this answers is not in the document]
bdeluga: Fragment was already introduced, in form of `<></>` but this way i couldn't set key, so i imported `Fragment` and used it. 🚀

mpiwonski (approved):

Alessiass (approved):
Looks fine, still we need to unify our prettier config because constant reformatting obfuscates code during review.

aniasgit (approved):

monika-klokosinska-intive (approved):

apps/storybook/stories/ButtonGroup.stories.tsx
```diff
@@ -18,7 +18,7 @@ const FullWidth: ComponentStory<typeof ButtonGroup> = (args) => (
   </div>
 );
 
-const FixedWidth: ComponentStory<typeof ButtonGroup> = (args) => (
+const FixedWidth: ComponentStory<typeof ButtonGroup> = args => (
```
wlechowicz, line 21: You have conflicting formatter settings. This will be fixed in the new config story, for now please use Prettier with those basic settings:
```
{
  "arrowParens": "always",
  "bracketSpacing": true,
  "endOfLine": "lf",
  "htmlWhitespaceSensitivity": "css",
  "insertPragma": false,
  "singleAttributePerLine": false,
  "bracketSameLine": false,
  "jsxBracketSameLine": false,
  "jsxSingleQuote": false,
  "printWidth": 80,
  "proseWrap": "preserve",
  "quoteProps": "as-needed",
  "requirePragma": false,
  "semi": true,
  "singleQuote": false,
  "tabWidth": 2,
  "trailingComma": "es5",
  "useTabs": false
}

File: packages/ui/ButtonGroup/index.tsx [hunk 3]
```diff
@@ -1,41 +1,41 @@
...
+import "material-symbols";
+import React from "react";
 
 export const ButtonGroup = ({ options, secondary }: GroupProps) => {
   return (
     <ButtonGroupStyled options={options} secondary={secondary}>
       {options.map(({ component, onSelect }, index) => {
         return (
-          <>
+          <Fragment key={`group-${index}`}>
```
wlechowicz, line 12: index-based key is almost always a bad idea. If you change the order, the elements will stay the same but with different keys, so not only everything will re-render, but animations will not work if you try to add something like autoanimate etc.

Also a key on Fragment makes little sense, that's why you were seeing an error in the first place, probably. You should wrap the input and the label in something real (an actual element) and put the key on that.

packages/ui/ButtonGroup/index.tsx
```diff
@@ -1,41 +1,41 @@
...
             <input
-              type={'radio'}
+              type={"radio"}
               id={`button-${index}`}
               key={`input-${index}`}
               onClick={onSelect}
-              name={'button-group'}
+              name={"button-group"}
             />
             <label key={`index-${index}`} htmlFor={`button-${index}`}>
```
wlechowicz, line 20: You don't need keys here, only on the parent.

wlechowicz (changes requested):

packages/ui/ButtonGroup/index.tsx
```diff
@@ -1,41 +1,41 @@
...
-import styled, { css } from 'styled-components'
-import 'material-symbols'
-import React from 'react'
+"use client";
+import { Fragment, ReactNode } from "react";
+import styled, { css } from "styled-components";
+import "material-symbols";
+import React from "react";
 
 export const ButtonGroup = ({ options, secondary }: GroupProps) => {
```
wyola, line 7: @bdeluga shouldn't this component have some option for a default value? right now nothing is checked at start and those button groups are used on pages like "Bills" where when user lands on it something should be selected so the content renders 

image

monika-klokosinska-intive (approved):

apps/storybook/stories/ButtonGroup.stories.tsx
bdeluga: Done.

packages/ui/ButtonGroup/index.tsx
bdeluga: Changed.

packages/ui/ButtonGroup/index.tsx
bdeluga: Added `defaultChecked?: boolean`, thanks !

packages/ui/ButtonGroup/index.tsx
```diff
@@ -127,4 +127,4 @@ const ButtonGroupStyled = styled.div<GroupProps>`
             color: #b1b1b1;
           `}
   }
-`
+`;
```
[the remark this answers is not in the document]
bdeluga: Done.

monika-klokosinska-intive (changes requested):

packages/ui/ButtonGroup/index.tsx
```diff
@@ -1,41 +1,40 @@
...
 
 export const ButtonGroup = ({ options, secondary }: GroupProps) => {
   return (
     <ButtonGroupStyled options={options} secondary={secondary}>
-      {options.map(({ component, onSelect }, index) => {
+      {options.map(({ component, onSelect, defaultChecked }, index) => {
         return (
-          <>
+          <Fragment key={`group-${index}`}>
             <input
```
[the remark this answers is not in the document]
bdeluga: Right! Thanks for pointing it out chief 👀

packages/ui/ButtonGroup/index.tsx
bdeluga: Problem was that Fragment is being rendered in `Array.map` and it needs key, even though it does not appear in DOM tree... I was using Fragment to avoid going deeper in css selector, it works fine imo.

About index-based key, i strongly agree, one solution could be requirement of id in single option object, what do you think about it ? 👀 @wlechowicz

bdeluga: Right! 🚀

bdeluga:
@monika-klokosinska-intive , @wlechowicz 
Did i resolved your requests ?

packages/ui/ButtonGroup/index.tsx
wlechowicz: Yeah, that sounds ok

apps/storybook/stories/ButtonGroup.stories.tsx
```diff
@@ -56,18 +56,21 @@ Primary.args = {
...
         console.log("Option 2 selected");
       },
+      id: "income",
     },
     {
       component: "Expenses",
       onSelect: () => {
         console.log("Option 3 selected");
       },
+      id: "expanses",
```
wlechowicz, line 73:
```suggestion
      id: "expenses",
```

apps/storybook/stories/ButtonGroup.stories.tsx
```diff
@@ -81,12 +84,15 @@ Secondary.args = {
...
       },
+      id: "income",
     },
     {
       component: "Expenses",
       onSelect: () => {
         console.log("Option 2 selected");
       },
+      defaultChecked: true,
+      id: "expanses",
```
wlechowicz, line 95:
```suggestion
      id: "expenses",
```

wlechowicz (changes requested):
Some typos to fix, rest looks good.

wlechowicz (approved):

PR #20, opened 2023-03-27
feat: welcome page #P23-70

zuzannapaw:


apps/web/app/SideNav.tsx
```diff
@@ -0,0 +1,22 @@
+"use client"
+import styled from "styled-components";
+import sidenav from  "../../../packages/ui/helper-imgs/sidenav.png"
```
bdeluga, line 3: I suggest moving all image files to `web/public` folder and just reference them either by `import` or by using `src="/your_image"`

apps/web/app/css/body.css
```diff
@@ -0,0 +1,4 @@
+body{
+    margin:0;
+    height:94.1vh;
```
bdeluga, line 3: This height of body creates vertical scroll, what are your thoughts behind it ? 😄

mpiwonski (changes requested):
I do agree with @bdeluga point regarding the need for web/public folder for images. I'll try to get this side navigation bar menu ready before you merge this branch, so you won't need this placeholder image for that component :)

apps/web/app/css/body.css
zuzannapaw: I had problems with adjusting body and html for whole welcome page (i couldn't handle the white gaps between body and html). But i've thought about replacing it by this:

`html{
    height: 100vh;
}

body{
    height: 94.1%;
}

html,body{
    margin: 0;
}
`
what do you think of that? The body is 94.1% because i need to left some space for top navigation (it does not belong to body).

apps/web/app/SideNav.tsx
zuzannapaw: Thanks! I will move it, of course it's only a temporal solution before having all components avaibles on main ☺️

zuzannapaw: Done! 🎉

apps/web/app/css/body.css
bdeluga: Everything belongs in body 😄, besides what navigation are we talking about ? 👀

zuzannapaw: Woops, yes... I previously checked mistakenly and now I can see it. Right, that's how usually it is, that everything belongs to body. Will work on that, thank you for noticing, that was important 💪

apps/web/app/layout.tsx
```diff
@@ -13,6 +15,16 @@ const inter = Inter({
   subsets: ["latin"],
 });
 
+const Main = styled.div`
+  display: flex;
+  justify-content: space-between;
+  height:100%;
```
aniasgit, line 21: With these settings it doesn't display well, on my screen the Card is higher than screen and there is scrollling. I suggest body height: 100% and here height: calc(100% - 68.88px), nav height is 68.88px

apps/web/app/welcome-page/page.tsx
```diff
@@ -0,0 +1,30 @@
+"use client"
+
+import styled from "styled-components"
+import { Card } from "ui";
+import "../css/body.css"
+
+const TypoStyled = styled.h1`
```
aniasgit, line 7: I suggest to put styled components under the main function.

aniasgit (changes requested):

apps/web/app/layout.tsx
zuzannapaw: Thank you, I will work on display! 👀

zuzannapaw: Thank you, I will fix it:)

apps/web/app/welcome-page/page.tsx
zuzannapaw: Do you mean to put for exemple `const TypoStyled = styled.h1 ...` under the `function WelcomePage()` ?

apps/web/app/layout.tsx
zuzannapaw: @wlechowicz I did some correction, how does it look now?

apps/web/app/SideNav.tsx
```diff
@@ -0,0 +1,22 @@
...
+`
+
+export default function SideNav() {
+    return (
+        <SideNavStyled>
+            <ImageWrapper><Image src={sidenav} width={120} height={400} alt="sidenav"></Image></ImageWrapper>
+        </SideNavStyled>
+    );
+  }
+
```
monika-klokosinska-intive, line 22: @zuzannapaw Inserting a new line automatically seems to have failed ;),

zuzannapaw: ooh :/ i will work it out 👁️

apps/web/app/Nav.tsx
```diff
@@ -1,43 +1,40 @@
...
+//Image will be replaced by Avatar
 export default function Nav() {
   return (
     <NavBar>
       <Logo white />
       <ActionWrapper>
-        <LinkStyled href="/">Home</LinkStyled>{" "}
-        <LinkStyled href="/users">Users</LinkStyled>
+        <IconWrapper><Icon icon={"notifications"} color={"white"}/></IconWrapper>
+        <Image src={avatar} width={30} height={30} alt="avatar"></Image>
```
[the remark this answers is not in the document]
zuzannapaw: I'll wait, thanks!

mpiwonski (approved):

apps/web/app/home/page.tsx
```diff
@@ -0,0 +1,29 @@
...
+    align-self: flex-start;
+    margin-top:23%;
+    font-family: "Inter", sans-serif;
+    font-size: 3.5em;
+    color: #1E4C40;
+`
+export default function WelcomePage() {
+    return (
+        <CardWrapperStyled>
+            <Card className="card-styled">
```
aniasgit, line 24: Instead of className you can use styled components,  for example 
```
const WelcomeCardStyled = styled(Card)`
  min-width: 95%;
  min-height: 95%;
`;
```
```
export default function WelcomePage() {
  return (
    <CardWrapperStyled>
      <WelcomeCardStyled>
        <TypoStyled>Welcome to Inbudget</TypoStyled>
      </WelcomeCardStyled>
    </CardWrapperStyled>
  );
```

zuzannapaw: Thank you, I will work on that, but I've thought that prop `className` was added to Card to use it like in situation above. Styled components would be better here?

wyola: @zuzannapaw  we add `className` prop to use it like @aniasgit suggested. When you do `const WelcomeCardStyled = styled(Card)` - it  passes new className, which is created dynamically, to Card

apps/web/app/css/global.css
```diff
@@ -0,0 +1,17 @@
...
+
+html,
+body {
+    margin: 0;
+}
+
+.card-styled {
+    min-width: 95%;
+    min-height: 95%;
+}
```
wyola, line 17: @zuzannapaw you can remove this piece of code when you apply suggested changes from here https://github.com/intive/patronage2023-js/pull/20/files#r1151822236 :)

apps/web/app/Nav.tsx
```diff
@@ -1,43 +1,40 @@
...
-  font-size: 1.5em;
-
-  &:hover {
-    text-decoration: underline;
-  }
+  justify-content: space-around;
+  align-items: center;
 `;
 
+const IconWrapper = styled.div`
```
Noshi96, line 25: You can style ```Icon``` itself within styled components same as @aniasgit  mentioned about ```Card``` ([PR comment](https://github.com/intive/patronage2023-js/pull/20/files#r1151822236)) or even remove ```IconWrapper```, since Icon is already centered.

apps/web/app/layouts/NavigationLayout.tsx
```diff
@@ -0,0 +1,26 @@
+"use client"
+
+import styled from "styled-components";
+import Nav from "../Nav";
+import SideNav from "../SideNavigationBar";
+import { LayoutProps } from "../layout";
+
+const Main = styled.div`
```
Noshi96, line 8: Since it is ```Main``` I would change tag to ```styled.main```
```suggestion
const Main = styled.main`
```

apps/web/app/home/page.tsx
zuzannapaw: @wyola  thank you for the explaination! that's very clever. ⭐ I will change it!

apps/web/app/layouts/NavigationLayout.tsx
zuzannapaw: That's a really good idea, I will change it!

apps/web/app/css/global.css
zuzannapaw: Yes, it makes sense now. On my way to change it 👍

apps/web/app/home/page.tsx
zuzannapaw: Done ☺️

monika-klokosinska-intive:
@zuzannapaw the vercel deployment for web is still broken on this branch - could you take a look and see what's the problem? This is the main merge blocker at the moment.

apps/web/app/Nav.tsx
```diff
@@ -1,43 +1,33 @@
...
 
 export default function Nav() {
   return (
     <NavBar>
       <Logo white />
       <ActionWrapper>
-        <LinkStyled href="/">Home</LinkStyled>{" "}
-        <LinkStyled href="/users">Users</LinkStyled>
+        <Icon icon={"notifications"} color={"white"}/>
+        <Avatar src="../public/avatar.png" outlined={true} />
```
monika-klokosinska-intive, line 30:
```suggestion
        <Avatar src="..avatar.png" outlined />
```
even! There is no need to pass true for a boolean attribute.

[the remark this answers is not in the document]
aniasgit: But this avatar displays very small, in Figma it's larger than the icon

wyola, line 30: use path `src="/avatar.png"` and it will work fine 👏 also, I would rather use SVG file, it can be exported from figma

aniasgit (approved):

AzulloCyll (approved):

aniasgit:
> @zuzannapaw the vercel deployment for web is still broken on this branch - could you take a look and see what's the problem? This is the main merge blocker at the moment.

I'm wondering what's wrong? When I run build on this branch locally I haven't any problems.

monika-klokosinska-intive (approved):
@zuzannapaw Looks good to me! Great job!

wlechowicz (approved):
LGTM

PR #22, opened 2023-03-29
feat: log-in form #P23-148

PawelPazderski:


apps/web/app/log-in/page.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const FormWrapper = styled.div`
+  margin: 0 auto;
+  padding-top: 10rem;
+  width: 416px;
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  margin: 0 auto;
```
monika-klokosinska-intive, line 17: @PawelPazderski double definition of `margin` in this class :)

PawelPazderski: Right, I'll fix that in a second :) thanks

apps/web/app/log-in/page.tsx
```diff
@@ -0,0 +1,93 @@
...
+  justify-content: space-between;
+  margin-bottom: 5rem;
+`;
+
+const ErrorWrapper = styled.div`
+  position: absolute;
+  top: 1rem;
+  left: 50%;
+  transform: translate(-50%, 0);
+  z-index: 5;
```
monika-klokosinska-intive, line 36: @PawelPazderski please add `width: 100%` here, it will make the error message as big as the inputs :)

PawelPazderski: done, thanks

apps/web/app/log-in/page.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return (
+      <Form
+        onSubmit={(values) => {
+          values.email === "smutnarzaba@png.pl" &&
+          values.password === "frytki123"
+            ? router.push("/")
+            : setErrMsg("Invalid credentials. Please try again.");
+        }}
+      >
```
Noshi96, line 56: Your form isn't actually an HTML ```<form>```, so we're losing some of the capabilities that come with using HTML form.
[House Form eg](https://houseform.dev/guides/html-forms.html) and [House Form](https://houseform.dev/introduction.html#example-usage) 
I think it's a good idea to add ```<form>``` to our form (good from an SEO perspective, I think)

apps/web/app/log-in/page.tsx
```diff
@@ -0,0 +1,94 @@
...
+              </Field>
+              <Field name="password">
+                {({ value, setValue }) => (
+                  <Input
+                    type="password"
+                    label="Password"
+                    value={value}
+                    onChange={(e) => setValue(e.currentTarget.value)}
+                    onFocus={closeErrorMessage}
+                  />
```
Noshi96, line 84: I'm not sure why it works that way, but if you click on the "Password" label when the input doesn't have focus, it won't open. You can solve this by adding the ```name="password"``` to the ```Input```

wlechowicz:
Please pull in changes from main once #18 is merged and put your form in the sign-in page there.

apps/web/app/log-in/page.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  return (
+      <Form
+        onSubmit={(values) => {
+          values.email === "smutnarzaba@png.pl" &&
+          values.password === "frytki123"
+            ? router.push("/")
+            : setErrMsg("Invalid credentials. Please try again.");
+        }}
```
Noshi96, line 55: I wondering about setting focus to email input when credentails are invalid

apps/web/app/log-in/page.tsx
wyola: @Noshi96 I think that @PawelPazderski is using here old version of the input, the updated one has name prop. Nevertheless, it is strange, since it was working well also in old version.

PawelPazderski: I added the name property works fine didn't see that thanks

apps/web/app/log-in/page.tsx
PawelPazderski: I wouldn't go that way there is a onFocus function that closes the Error Message

PawelPazderski:
> Please pull in changes from main once #18 is merged and put your form in the sign-in page there.

Sure, made small fixes I'm waiting for the layout merge

apps/web/app/log-in/page.tsx
Noshi96: What I mean is that when you use tabs and spaces to navigate through the application right after validation, in your case the user has to go through all the page elements again. I've seen the behavior with focusing on incorrect input data in Facebook and Spotify login forms, for example

apps/web/app/log-in/page.tsx
```diff
@@ -0,0 +1,96 @@
...
+  justify-content: space-between;
+  margin-bottom: 5rem;
+`;
+
+const ErrorWrapper = styled.div`
+  position: absolute;
+  width: 100%;
+  top: 1rem;
+  left: 50%;
+  transform: translate(-50%, 0);
```
aniasgit, line 36: If you translate only along x axis you can use transform: translateX(-50%)

apps/web/app/log-in/page.tsx
```diff
@@ -0,0 +1,96 @@
...
+  z-index: 5;
+`;
+
+export default function LogInPage() {
+  const router = useRouter();
+  const [errMsg, setErrMsg] = useState("");
+
+  const closeErrorMessage = (): void => {
```
aniasgit, line 44: You don't need this void here.

apps/web/app/log-in/page.tsx
aniasgit: I agree with @Noshi96.

apps/web/app/log-in/page.tsx
PawelPazderski: Yes I can :) let's change that

apps/web/app/log-in/page.tsx
PawelPazderski: Right it works fine without still a bit confused if it comes to ts

apps/web/app/log-in/page.tsx
PawelPazderski: Let's leave that to discuss on a meeting I think, for what I can tell there will be changes neccessary to do in input component ( the button show/hide password submits the form for example )

PawelPazderski: I had a chat with Wojtek about it, you're right the form should be added

monika-klokosinska-intive:
@PawelPazderski Looking good 👍 ! Once the layout is merged and used, this should be good to go!

aniasgit (approved):
Looks good for me 👍

monika-klokosinska-intive (approved):
@PawelPazderski  looks good to me!

Noshi96 (approved):

crutchcorn:
Hey, y'all! I'm the original author of HouseForm.

First off: Apologies for reaching out in this way. I couldn't think of a more appropriate avenue to reach out. 😅

As far as I'm aware, you're all the first folks that have used HouseForm on a company GitHub organization! That's legitimately so exciting! 

I also noticed links to the docs for native HTML form interop, which makes me super glad I spent the time to write the page. 😄

Anyway - to the point - I was wondering if:

1) Do you had any feedback on using HouseForm? The good, the bad, and anything in between. The project is early, and I'm hungry for feedback to improve the project.

2) Are you considering using HouseForm in non-OSS production? If so, is there anything I can do to help unblock the usage? I'd be happy to do an initial call, pair programming, or bug prioritization. Not looking to gain financial sponsorship out of this particular interaction (although I wouldn't complain 😛), but simply hoping to help push the adoption of the library forward and understand your use case.

Either way, I hope y'all are doing great! Apologies for the disruption.

wlechowicz:
Hey @crutchcorn , great to have you here!

First off, this project is part of intive's Patronage, 2023 edition, which is an internship program of sorts - we get applications from students or people wanting to enter IT, we pick the best talent, build a team, assign leaders and lead them through a fully-fledged project over a number of months (design, product owners, demos, the lot). The apps we create in the process are not commercial and they are always public repos. I also strive to make the learning path fun for the participants, which I'm sure you vibe with.

One of the tasks that we did couple of sprints ago was an evaluation of React form libraries, where assigned team members - one per library - had to prepare a list of pros and cons and a POC of a given library. We evaluated Formik, react-hook-form and react-final-form. None of those really fit our use case well, the best of the bunch being Formik, which I've used commercially a lot and I was (too) aware of its shortcomings. That's when I threw HouseForm into the pool, which I picked up from Theo's community.
It happened to perfectly balance the shortcomings of other libs while delivering exactly what we needed: help us validate the form and collect its data for sending, without being too much of an overhead or a bloat of code. We'd also love a not-forced integration with zod.

For me, as a leader of the project, the apps we create during Patronage annual editions often serve as a playground and/or proving grounds for tech that I then transfer over to my commercial projects. I assume if a piece of tech - say a library or a pattern - is easily understood and successfully applied by Patronage's pre-juniors, then it's on a good path to be a success in a "real" project with seniors. One such example could be a custom reducer-based action solution that I introduced last year and have scaled it up since with great success in "serious" projects. HouseForm also fits the bill so far, but we're only just starting our BE integrations so it's too early for any specific feedback. It has been smooth sailing so far but it might still be a case of rose-tinted glasses. I'd be happy to share feedback with you in a couple of sprints.

If you happen to have any tips on how we (mis)used HouseForm, I'd appreciate greatly if you shared. Thanks for coming by!

crutchcorn:
That's an awesome process for helping mentor and guide others in tech. Love to hear about this program.

I think your reasons for picking HouseForm are super well-aligned with the rationale for why I built HouseForm 😄 I, too, had the most experience with Formik prior to building it (we used it at Hilton a couple of years ago) and enjoyed its usage but saw some of the downsides, especially as of late.

> For me, as a leader of the project, the apps we create during Patronage annual editions often serve as a playground and/or proving grounds for tech that I then transfer over to my commercial projects.

Love this - what a great feedback loop and great thought process with regard to your Junior-first methodology.

> I'd be happy to share feedback with you in a couple of sprints.

Please do! I'm available anywhere on GitHub ([maybe a GH Discussion on HouseForm would fit well?](https://github.com/houseform/houseform/discussions)) or, alternatively, [I can be DM'd on Twitter](https://twitter.com/crutchcorn), or even [our official HouseForm Discord.](https://discord.gg/n9pGcTh3Sx)

> If you happen to have any tips on how we (mis)used HouseForm

I've reviewed all of the usages of HouseForm in the current codebase - y'all have done incredible work with it!

A few small thoughts:

1) In `PasswordSubComponent`, you have a rather lengthy RegEx here. I might break this out to be outside of the component declaration. While having your validation logic inline is a boon for DX, it can bloat our UI files. I like to think that even having the reference to the validation in the UI layer helps a lot with DX as well.

https://github.com/intive/patronage2023-js/blob/60b6097d4b6bb79d7edeadd1637fe8c62778dbb3/apps/web/app/(regflow)/sign-up/PasswordSubComponent.tsx#L95-L111

2) On that same page, I think there might be some room for [an optimization that's built right into HouseForm - `memoChild`.](https://houseform.dev/guides/performance-optimizations.html)

    This optimization can be a bit advanced for juniors, but the gist of it is that if you have data changing outside of `Form` regularly, the `Form` will do repeated and expensive re-renders that you cannot opt out of without using `useMemo` or our helper `memoChild`.
    
    
     - If you do want to explore using `memoChild` but worried about the educational aspect of the tool, [I've written a blog post that explains it's internals and the need to use a similar optimization even with other libraries 😊](https://unicorn-utterances.com/posts/functions-are-killing-react-performance#children-functions)

     - I'm absolutely not going to lie - the DX for `memoChild` could be improved. It's stable with a fair number of tests backing it, but we don't have the `"Click to auto-fill the dependency array"`, for example. [I'm working on an ESLint rule to fix this problem](https://github.com/houseform/houseform/pull/72), but it might not be ready right away.

All in all, keep up the great work y'all, and thanks for the reply!

PR #23, opened 2023-03-29
feat: add Prettier config #P23-162

Vectari:
Here is my proposition of Prettier configuration. If you have any suggestion about rules (.prettierrc) or ignore (.prettierignore) just let me know.
Make sure that you have checked settings in VS Code:

1. Editor: Format On Save
2. Editor: Default Formatter -> Prettier - Code formatter

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,7 +1,7 @@
 {
   "semi": true,
   "tabWidth": 2,
-  "singleQuote": true,
+  "singleQuote": false,
   "trailingComma": "es5",
   "bracketSpacing": true,
   "bracketSameLine": true,
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,8 @@
+{
+  "semi": true,
+  "tabWidth": 2,
+  "singleQuote": true,
+  "trailingComma": "es5",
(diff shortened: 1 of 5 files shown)


.prettierrc
```diff
@@ -0,0 +1,15 @@
+{
+  "semi": true,
+  "tabWidth": 2,
+  "singleQuote": true,
```
aniasgit, line 4: Do we want a `singleQuote`? Until now, double quote was used almost everywhere.

Vectari: I use double too but that's how I understood Wojtek's suggestions. Maybe I understand it wrong...

aniasgit: Look here https://github.com/intive/patronage2023-js/pull/19#discussion_r1152118840

Vectari: Yes, I seen it. Maybe I understand something wrong. I fix it after work.

Vectari: On phone I can too 🙂

wlechowicz (approved):

wyola (approved):
looks good to me

aniasgit (approved):

PR #24, opened 2023-03-30
fix: delete unnecessary exports and story #P23-124

mpiwonski:


wlechowicz (approved):
This resolves build-time error:
```
storybook:build: Export "SideNavigationBarItem" of module "../../packages/ui/SideNavigationBarItem/index.tsx" was reexported through module "../../packages/ui/index.tsx" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
storybook:build: Either change the import in "stories/SideNavigationBar.stories.tsx" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
```

monika-klokosinska-intive (approved):

PR #28, opened 2023-03-31
feat: #P23-165 avatar-selector

bdeluga:


packages/ui/AvatarSelector/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  overflow-y: hidden;
+  padding: 2rem;
+  gap: 1rem;
+
+  /* Hide scrollbar for Chrome, Safari and Opera */
+  &::-webkit-scrollbar {
+    display: none;
+  }
```
PawelPazderski, line 51: Might consider leaving a scrollbar here (probably would need some styling) and hide it for touchscreens what is I assume the idea here, at this point while testing the component it's almost impossible to scroll it if not in mobile mode below the breakpoint. I found a solution for that, leaves a scroll on a standard use in browser and hides it while in mobile view - it's just a suggestion, I never used it before and can't be sure if it is 100% ok so I'll just leave it here to discuss
```
@media (hover: none) {
    &::-webkit-scrollbar {
    display: none;
    }
  }
```

PawelPazderski (commented):
Looks very nice 👍

Noshi96 (commented):
I really like both the look and the easy-to-read code! @PawelPazderski made a good point about the issue with sliding on a shrank desktop view, and I agree that we need some sort of scroll bar, even if we can provide changing avatars by clicking tab.

packages/ui/AvatarSelector/index.tsx
bdeluga: Sure, i'll get to work then, thanks for pointing it out ! 👀

apps/web/app/AvatarSelectorDemo.tsx
```diff
@@ -0,0 +1,56 @@
+"use client";
+import React from "react";
+import { AvatarSelector } from "ui";
+export default function AvatarSelectorDemo() {
+  const [selectedAvatar, setSelectedAvatar] = React.useState<string>();
+
+  return (
+    <div
+      style={{
```
AzulloCyll, line 9: Szhould be changet to 100% maybe.

packages/ui/AvatarSelector/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+const ButtonStyled = styled.button<ButtonProps>`
+  cursor: pointer;
+  border: none;
+  font-size: 88px;
+  height: 1em;
+  padding: 0;
+  background: none;
+  border-radius: 50%;
+  outline-offset: 2px;
+  outline: ${({ selected }) => (selected ? "2px solid #000" : "none")};
```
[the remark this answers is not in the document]
aniasgit: I agree with @AzulloCyll

apps/web/app/AvatarSelectorDemo.tsx
```diff
@@ -0,0 +1,56 @@
...
+        width: "480px",
+      }}
```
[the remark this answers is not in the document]
bdeluga: The whole 480px thing is to showcase that component has width of 100%.I had to manally set correct width to make it appear like in Figma.

packages/ui/AvatarSelector/index.tsx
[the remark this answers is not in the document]
bdeluga: I haven't merged main in ages, so I will take that into account. Without that comment, I wouldn't have thought about it.
Thanks 😄

AzulloCyll (approved):

monika-klokosinska-intive:
@bdeluga please resolve the conflicts! Except that it looks good to me!

apps/web/app/AvatarSelectorDemo.tsx
```diff
@@ -0,0 +1,56 @@
+"use client";
+import React from "react";
```
aniasgit, line 2: You don't need import React

aniasgit, line 2:
```suggestion
import React from "react";
import {useState} from "react"
```

apps/web/app/AvatarSelectorDemo.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { AvatarSelector } from "ui";
+export default function AvatarSelectorDemo() {
+  const [selectedAvatar, setSelectedAvatar] = React.useState<string>();
```
aniasgit, line 5:
```suggestion
  const [selectedAvatar, setSelectedAvatar] = useState<string>();
```

wlechowicz:
It doesn't look right in Storybook. If you switch to mobile view, the scroll bar is cut off and the white gradient is visible on the left, instead of on the right edge.

bdeluga:
> It doesn't look right in Storybook. If you switch to mobile view, the scroll bar is cut off and the white gradient is visible on the left, instead of on the right edge.

I added gradient to both sides, so it doesn't look strange with gradient only on right side of component. Scrollbar gets cuts off, which i'll fix asap.

apps/web/app/AvatarSelectorDemo.tsx
bdeluga: Its from old demo ;)

bdeluga:
Marging component, so it doesn't block @PawelPazderski [Task](https://tracker.intive.com/jira/browse/PATRO23-209). In case of bugs I'll resolve them ASAP.

wlechowicz:
@bdeluga no it does not, gradient is only present on left side and scrollbar is cut off. If the list grows longer, the scrollbar will go well outside the screen.

image

PR #29, opened 2023-04-02
ref: improve rwd of registration layout #P23-198

AzulloCyll:


apps/web/app/(regflow)/sign-in/layout.styles.tsx
```diff
@@ -0,0 +1,83 @@
...
+    flex-direction: column;
+    justify-content: center;
+    padding-left: 6%;
+    padding-right: 3%;
+  }
+`;
+
+export const FormWrapperStyled = styled.div`
+  overflow-x: auto;
+  padding: 8px;
```
wyola, line 48: padding is very small, especially on the right side, while there is a lot of space between h1 and the form wrapper

image

AzulloCyll: Thanks for your review. Changed this padding to 16px. But many other changes was added including better example of sign-up page.

apps/web/app/(regflow)/layout.styles.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  ${device.desktop} {
+    margin-top: 32px;
+    font-size: 2.5rem;
+    line-height: 2.5rem;
+  }
+`;
+
+export const ParagraphStyled = styled.p`
+  font-family: "Inter", sans-serif;
```
aniasgit, line 71: We have set Inter as the default font in our RootLayout, so there is no need to add font-family here. Especially that in this form it doesn't work and Arial is rendered (for me) and not Inter.

apps/web/app/(regflow)/layout.styles.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  ${device.desktop} {
+    font-size: 1rem;
+    line-height: 1.5rem;
+    margin-top: 1rem;
+  }
+`;
+
+export const CustomCard = styled(Card)`
+  font-family: "Inter", sans-serif;
```
aniasgit, line 85: same as here: https://github.com/intive/patronage2023-js/pull/29/files#r1155777752

wyola (commented):
In this view, I would make the card the same width as the h1, but it is just my personal preference and I don't know what was decided about it during the layout call

image

apps/web/app/(regflow)/layout.styles.tsx
aniasgit: after removing font-family, inter is render now :)

apps/web/lib/media-queries.tsx
```diff
@@ -0,0 +1,9 @@
+const size = {
+  desktop: '768px',
+  large: '1200px'
+};
+
+export const device = {
+  desktop: `@media (min-width: ${size.desktop})`,
+  large: `@media (min-width: ${size.large})`,
+};
```
wyola, line 9: add empty line at the end

monika-klokosinska-intive:
Screenshot 2023-04-04 at 18 06 12
@AzulloCyll I am getting a weird behaviour around 1020px width - a horizontal scroll or a cut down view. Could you take a look?

monika-klokosinska-intive:
@AzulloCyll nevermind, I am not able to reproduce. Let's chalk it up to my local's weirdness and leave it as it!

aniasgit (approved):
It looks good to me 👍

monika-klokosinska-intive (approved):

PR #30, opened 2023-04-03
fix: improve rwd of navigation layout #P23-199

zuzannapaw:
### Important changes:
 `(multiCard)` folder does not already exists. Application can still have `(singleCard)` layout, and for pages with two cards, a component `<MultiCardLayout>` must be imported and it must accept arguments `main` - for bigger, left card-  and `aside` for smaller right card. 

`<DummyAsideCardContent>` was created just for now as a dummy component passed as `aside` prop in `<MultiCardLayout>` for a proper presentation of a layout that has two cards - this is not an official component.


apps/web/app/(navigation)/(multiCard)/layout.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const RightCardStyled = styled(CardStyled)`
+  display: none;
+  @media ${device.desktop} {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    align-self: flex-start;
+    flex-basis: 15%;
+    min-width: 300px;
```
wyola, line 16: I think that the width of the right card should change a bit with the screen since on smaller screens it takes up half of the space. The card on the left needs to display a lot of things and it can be difficult when there will be left for it ~370px.

image

zuzannapaw: Thank you, I will work on that! 💪

apps/web/app/(navigation)/(multiCard)/layout.tsx
```diff
@@ -0,0 +1,81 @@
...
+    min-height: 250px;
+    margin-left: 1%;
```
Noshi96, line 18: I think it is better to change it to 2% (the padding value of the parent) and keep the same spaces around the cards. Or rather, use ```gap: 2%;``` in ```CardWrapperStyled``` instead of the margin here ;D
```suggestion
    margin-left: 2%;
```
image

AzulloCyll: I was tought that % and vh, vv used in RWD layouts is a big no no.
So i recommend to set margins to constant values.

zuzannapaw: @Noshi96 great advice,thanks! Is that a rule only for margins and paddings?

apps/web/app/css/global.css
```diff
@@ -7,5 +7,7 @@
 body,
 html {
   width: 100%;
+  min-width: 300px;
```
monika-klokosinska-intive, line 10: @zuzannapaw  Are these necessary? As long as the pages have content they should always have some width as well. What was the reason you decided to go for `min-width` and `min-height` here?

zuzannapaw: @monika-klokosinska-intive when html & body don't have min-width and min-height, the whole page doesn't stop narrowing down, so I thought a good idea would be to add min-width and min-height to prevent this behavior.
image

apps/web/app/(navigation)/(multiCard)/layout.tsx
zuzannapaw: okay it's done, I think it looks better and changing desktop dimensions for 1024px helped a lot. If the new breakpoint will appear (768px), I will adjust RightCard for it.

apps/web/app/(navigation)/(multiCard)/layout.tsx
```diff
@@ -0,0 +1,81 @@
...
+`;
+
+//creating imitation of RightCard content
+const DescriptionsWrapper = styled.div`
+  display: flex;
+  width: 100%;
+  justify-content: space-between;
+  align-items: center;
+  padding: 10px 20px 0 20px;
+  margin-top: 5%;
```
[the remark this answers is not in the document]
zuzannapaw: it's done, thank you for great advice ☺️

apps/web/app/(navigation)/(multiCard)/layout.tsx
zuzannapaw: @Noshi96 it's done ☺️

apps/web/app/css/global.css
monika-klokosinska-intive: @zuzannapaw I see what you mean! The page indeed doesn't break  with values like `min-width` present, but instead we get the vertical scrollbar visible which is also not ideal.
The smallest width we are currently supporting is 375px, so I would propose removing these min values anyway. If we ever need to support even smaller screens we will have to think of a different mobile view for them anyway.
Having a min-width and min-height on `html` might cause us some confusion down the line when debugging.

packages/ui/SideNavigationBar/index.tsx
```diff
@@ -33,7 +33,7 @@ const SideNavigationBarStyled = styled.ul<SideNavigationBarProps>`
   flex-direction: column;
   justify-content: flex-start;
   align-items: center;
-
+  margin-top: 0.5%;
```
monika-klokosinska-intive, line 36: @zuzannapaw please use either `px` or `em`s for this! Since this looks like a small margin, I think `px` might be fine in this case.

apps/web/app/(navigation)/(singleCard)/page.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const metadata: Metadata = {
+  title: "Home",
+  description: "Welcome to my page",
+};
+
+export default function MainPage() {
+  return (
+    <>
+      <TypoStyled>Welcome to Inbudget</TypoStyled>
```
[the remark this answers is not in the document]
aniasgit: @zuzannapaw and due to dictionary usage, you need to extract elements with text to client-side components

packages/ui/SideNavigationBar/index.tsx
zuzannapaw: Ohh, it's my oversight because it wasn't my component. Should I change it anyway?

apps/web/app/(navigation)/(singleCard)/page.tsx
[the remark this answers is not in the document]
zuzannapaw: Thank you, on my way to implement that! 💪

apps/web/app/css/global.css
zuzannapaw: okay! it's done!

packages/ui/SideNavigationBar/index.tsx
monika-klokosinska-intive: @zuzannapaw Please do!

AzulloCyll (approved):
Still wondering how left side navigation behave on smallest mobile phones.

But at this point i't works fine for me.  
Good job. 👍🏻

wlechowicz (approved):
Looks nice. We'll do any adjustments when we start implementing those screens that need the multi card layout.

packages/ui/SideNavigationBar/index.tsx
monika-klokosinska-intive: @zuzannapaw I still see percentages for margin ;). Please change even if you it's not your code!

apps/web/app/(navigation)/HomePageComponents.tsx
```diff
@@ -10,23 +11,32 @@ export const CardWrapperStyled = styled.div`
...
   color: #1e4c40;
+  margin-top: 35px;
+  @media ${device.desktop} {
+    font-size: 2.3em;
+  }
 `;
 export const TypoStyled = styled.h1`
   font-family: "Signika", sans-serif;
-  font-size: 3.5em;
+  font-size: 3em;
```
aniasgit, line 36: welcome page
Welcome page doesn't look good on the smallest device (375px), I think `LinkStyled` and `TypoStyled` should have some padding/margin. Also think about smaller font-size for mobiles

aniasgit (approved):
Resolve conflicts and it looks great 👍

apps/web/app/(navigation)/HomePageComponents.tsx
zuzannapaw: @aniasgit thank you for the remark, I have applied your suggestions! ☺️

packages/ui/SideNavigationBar/index.tsx
zuzannapaw: thank you, it's done

PR #31, opened 2023-04-03
feat: dictionary #P23-163

aniasgit:
In libs/dictionary.tsx exists the dictionary object, where every text should be added. We use texts in client-side components using the useTransalte hook. Usage example: 

image


apps/web/lib/contexts/LanguageContext.tsx
```diff
@@ -0,0 +1,25 @@
...
+  setLang: (newLanguage: string) => void;
+}
+
+export const LanguageContext = createContext<LanguageContextInterface>(
+  undefined!
+);
+
+export const LanguageProvider = ({ children }: any) => {
+  const [lang, setLang] = useState("en");
+
```
Noshi96, line 14:
```suggestion
interface LanguageContextInterface {
  currentLang: string;
  setLang: Dispatch<SetStateAction<string>>;
}

export const LanguageContext = createContext<LanguageContextInterface>(
  currentLang: "",
  setLang: () => {},
);

export const LanguageProvider = ({ children }: React.ReactNode) => {
  const [lang, setLang] = useState("en");

```
I'm thinking of a fixed, more specific type for setLang, and also initialize the context with the right properties, not undefined!

Noshi96 (approved):
Great job! :D

Alessiass (approved):
Good work! Looks really nice!

PR #32, opened 2023-04-03
refactor: input component improvements #P23-200

wyola:
- improvements discussed on slack
   - accessibility of error messages, 
   - prevent form submitting on icon button click
   - reserve additional space for supporting label so that inputs does not shift when stacked in the form
- slight animation delay was added to eliminate label jumping when the user clicks on the icon button while the password input is empty ⬇️ 

https://user-images.githubusercontent.com/80450315/229596033-6d96daaf-4a6b-4c9e-9882-996c60971f54.mp4

### DIFF
--- a/packages/ui/Input/index.tsx
+++ b/packages/ui/Input/index.tsx
@@ -2,11 +2,16 @@
 
 import { useState, useRef, useId } from "react";
 import styled, { css } from "styled-components";
+import { boolean } from "zod";
 
 import { Icon } from "../Icon";
 
 type StyledErrorProps = {
   hasError?: boolean;
+};
+
+type StyledInputProps = {
+  hasSupportingLabel?: boolean;
 };
 
 export type InputProps = {
@@ -86,6 +91,7 @@ export const Input = ({
         onChange={onChange}
         onFocus={onFocus}
         onBlur={onBlur}
+        hasSupportingLabel={Boolean(supportingLabel)}
       />
       <StyledLabel hasError={hasError} htmlFor={id || name}>
         {label}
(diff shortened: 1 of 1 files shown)


bdeluga (approved):
Looks good 🚀

Alessiass (approved):
Good change!

packages/ui/Input/index.tsx
```diff
@@ -38,10 +42,10 @@ export const Input = ({
     if (type === "password") {
       return (
         <StyledIcon
-          onClick={(e) => {
-            e.preventDefault();
-            setTypeOverride(typeOverride ? "" : "text");
+          onClick={(event) => {
+            event.preventDefault();
```
Noshi96, line 46: When we type something and click on the eye icon, focus sets our cursor to the beginning of the input. When we use stopPropagation() the cursor stays where we left it. TBH I don't know if this is a good approach. E.g. when we click show password on the Netflix form, the cursor behaves as it does in our application.
```suggestion
            event.stopPropagation();
```

Noshi96 (approved):

monika-klokosinska-intive (approved):

packages/ui/Input/index.tsx
wyola: `event.preventDefault()` was used to stop form submission when clicking icon button https://intive-apu8934.slack.com/archives/C04R6LM4JSV/p1680170725637589?thread_ts=1680170425.499899&cid=C04R6LM4JSV

from what I'm reading, `stopPropagation()` does not stop default behaviours

PR #34, opened 2023-04-05
feat: textarea component #P23-213

zuzannapaw:


packages/ui/Textarea/index.tsx
```diff
@@ -0,0 +1,22 @@
+"use client";
+import styled from "styled-components";
+
+const TextAreaStyled = styled.textarea`
+  width: 80%;
```
monika-klokosinska-intive, line 5: @zuzannapaw I would change this to `100%` - let the textarena take the whole parent element by default. This will make it simpler for anybody using the component - they won't have to calculate how much `80%` of the current width is.

packages/ui/Textarea/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  height: auto;
```
monika-klokosinska-intive, line 6: @zuzannapaw I think you can safely remove this.

packages/ui/Textarea/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  min-height: 70px;
+  font-family: "Inter", sans-serif;
+  font-size: 1em;
+  border-radius: 10px;
+  border: 2px solid ${({ theme }) => theme.textarea.disabled};
+  padding: 12px;
+  resize: none;
+
+  &:focus {
+    outline: 2px solid ${({ theme }) => theme.textarea.focus};
```
monika-klokosinska-intive, line 16: @zuzannapaw I would also suggest adding `border: 0;` to the focus styles. Right now the border is still visible together with the outline giving it a weird greyish-green look on focus.

packages/ui/Textarea/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  }
+`;
+
+export const Textarea = () => {
+  return <TextAreaStyled placeholder="Description" />;
```
monika-klokosinska-intive, line 21: @zuzannapaw Placeholder should be received as a prop - it might be something else than description on a different screen after all!

monika-klokosinska-intive: @zuzannapaw we will also need an `onChange` handler that calls a function received in props - similar to inputs. Otherwise the textarea has no way to communicate with the rest of the form.

monika-klokosinska-intive (changes requested):
@zuzannapaw Please take a look at my suggestions!

packages/ui/Textarea/index.tsx
monika-klokosinska-intive: @zuzannapaw and since we are adding props, please add `className` as well and pass it to styled component, so the styles can be extended. Some forms might want a bigger / smaller textarea and they should be able to overwrite the styles to get it :).

zuzannapaw: Thank you, I will apply all suggestions!

zuzannapaw: I think I have applied all your suggestions ☺️

monika-klokosinska-intive (approved):

aniasgit (approved):
It would be nice if the placeholder behaved like in input.

monika-klokosinska-intive (changes requested):
@zuzannapaw Please add the moving label behaviour as seen on inputs - you can use Ola's input code as a guide.

bdeluga (changes requested):
Please take a look at my suggestion 👀 
Other than that component is very nice ! 😄

packages/ui/Textarea/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  font-family: "Inter", sans-serif;
+  font-size: 1em;
+  border-radius: 10px;
+  border: 2px solid ${({ theme }) => theme.textarea.disabled};
+  padding: 12px;
+  resize: none;
+
+  &:focus {
+    border: 0;
+    outline: 2px solid ${({ theme }) => theme.textarea.focus};
```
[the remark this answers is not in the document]
zuzannapaw: @bdeluga good eye! I fixed it! ;)

monika-klokosinska-intive (approved):

monika-klokosinska-intive:
@zuzannapaw please resolve the conflicts and you are good to go!

bdeluga (approved):
Approved, good job! 🚀

PR #35, opened 2023-04-05
feat: icon picker component #P23-215

aniasgit:


packages/ui/IconPicker/iconPicker.styled.tsx
```diff
@@ -0,0 +1,77 @@
...
+  grid-template-columns: repeat(3, 64px);
+  grid-template-rows: repeat(3, 64px);
+  width: fit-content;
+  height: fit-content;
+  padding: 4px;
+`;
+
+export const SelectIconButtonStyled = styled.button`
+  display: flex;
+  justify-content: center;
```
Noshi96, line 62: You can add ```cursor: pointer;``` also for internal icon buttons

aniasgit: You're right 😄

packages/ui/IconPicker/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+  IconPickerStyled,
+  IconAndButtonWrapperStyled,
+  IconsSelectorStyled,
+  EditButtonStyled,
+  SelectIconButtonStyled,
+} from "./iconPicker.styled";
+import { Icon, IconType } from "../Icon";
+
+type IconPickerProps = {
+  defaultIcon?: IconType;
```
Noshi96, line 13: I think it is better to make the defaultIcon as a mandatory property.

monika-klokosinska-intive: @Noshi96 @aniasgit as the creator of this task I assumed there might be a use case where the user hasn't selected anything yet and we want to reflect that on the UI.

apps/storybook/stories/IconPicker.stories.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from "react";
```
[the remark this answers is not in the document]
aniasgit: If we don't import React in storybook, there are some problems: 
Bez tytułu2
Everybody use import React in storybook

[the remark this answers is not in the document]
monika-klokosinska-intive: @aniasgit Makes sense, thanks for the clarification!

monika-klokosinska-intive (approved):

packages/ui/IconPicker/iconPicker.styled.tsx
aniasgit: it's done

Noshi96 (approved):

Noshi96 (approved):

bdeluga (approved):
Looks, very good ! 👀

monika-klokosinska-intive:
@aniasgit I have one accessibility improvement to make - when the picker element with 9 icons is visible, pressing escape on the keyboard should close the picker. 

Great work otherwise!

aniasgit:
> @aniasgit I have one accessibility improvement to make - when the picker element with 9 icons is visible, pressing escape on the keyboard should close the picker.
> 
> Great work otherwise!

Good idea! I've just added it :)

monika-klokosinska-intive (approved):

PR #36, opened 2023-04-05
feat: sign up success-error screen #P23-210

Noshi96:
- Ticket: https://tracker.intive.com/jira/browse/PATRO23-210
- [Success Preview](https://patronage2023-js-web-git-feature-p23-210-step-f918d5-wlechowicz.vercel.app/sign-up)
- [Error Preview](https://patronage2023-js-web-git-feature-p23-210-step-f918d5-wlechowicz.vercel.app/sign-up/error)

Task: Prepare (4/4) screen for ```FlowController```.

```SuccessErrorScreen``` receives three types of props:
- ```success```: status of validation (```boolean```)
- ```onBackToStart```: function that takes the user back to the first screen in the registration process
- ```loginHref```: if the ```success``` status is true, the button on the screen will become a link with a href property equal to  ```loginHref```

in code:
```javascript 
<SuccessErrorScreen success={false} onBackToStart={onBackToStart} loginHref="/home" />
```

Changes:
- ```dictionary.tsx```: Added translations for SuccessErrorScreen elements
- ```theme.tsx```: Added colors to the theme object
- ```Icon/index.tsx```: Two new icons added

Fixes: 
- ```<LinkComponent>```: 
  - ```a``` was changed for Next ```Link```
  - can now be styled using styled components
- ```<Icon>```: 
  - can now be styled using styled components
 - ```<Button>```: 
   - can now be styled using styled components
 
To do:
- ~~wait for ```ref: improve rwd of registration layout #P23-198```  PR~~
- Adjust styles in the ```Card```

### DIFF
--- a/apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
+++ b/apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
@@ -1,0 +1,118 @@
+"use client";
+
+import { useTranslate } from "lib/hooks";
+import styled from "styled-components";
+import { Icon, LinkComponent, ButtonStyled } from "ui";
+
+type SuccessType = {
+  success: boolean;
+};
+type SuccessErrorScreenProps = {
+  onBackToStart: () => void;
+  loginHref: string;
+} & SuccessType &
+  React.HTMLProps<HTMLDivElement>;
+
+const ScreenContent = styled.div`
+  box-sizing: border-box;
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  flex-direction: column;
+  min-height: 518px;
+
+  @media only screen and (min-width: 1024px) {
+    justify-content: center;
+  }
+`;
+
+const ScreenStatusWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+`;
+
+const ScreenCircle = styled.div<SuccessType>`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 120px;
+  height: 120px;
+  border-radius: 50%;
+  background: ${({ success, theme }) =>
+    success ? theme.signUp.main : theme.signUp.error};
+`;
+
+const IconStyled = styled(Icon)`
+  color: ${({ theme }) => theme.signUp.icon};
+  font-weight: 700;
+`;
+
+const ScreenHeader = styled.h2`
+  font-size: 1.5em;
+  line-height: 1.5em;
+  font-family: "Signika", sans-serif;
+  margin-top: 32px;
+  color: ${({ theme }) => theme.signUp.main};
+`;
+
+const ScreenSubheader = styled.p`
+  line-height: 1.5em;
+  font-family: "Inter", sans-serif;
+  margin-top: 4px;
+  color: ${({ theme }) => theme.signUp.text};
+`;
+
+const LinkStyled = styled(LinkComponent)`
+  margin-top: 136px;
+  text-decoration: none;
+  border: 2px solid ${({ theme }) => theme.button.primary.main};
+  color: ${({ theme }) => theme.signUp.link};
+  background-color: ${({ theme }) => theme.button.primary.main};
+
+  @media only screen and (min-width: 1024px) {
+    margin-top: 96px;
+  }
+`;
+
+export const SuccessErrorScreen = ({
+  success,
+  onBackToStart,
+  loginHref,
+}: SuccessErrorScreenProps) => {
+  const { dict, t } = useTranslate("SignUpPage");
+  const {
+    successErrorScreen: { status },
+  } = dict;
+
+  const statusHeaderText = success
+    ? t(status.success.header)
+    : t(status.error.header);
+  const statusSubheaderText = success
+    ? t(status.success.subheader)
+    : t(status.error.subheader);
+  const statusButtonText = success
+    ? t(status.success.button)
+    : t(status.error.button);
+
+  return (
+    <ScreenContent>
+      <ScreenStatusWrapper>
+        <ScreenCircle success={success}>
+          <IconStyled icon={success ? "done" : "priority_high"} iconSize={56} />
+        </ScreenCircle>
+        <ScreenHeader>{statusHeaderText}</ScreenHeader>
+        <ScreenSubheader>{statusSubheaderText}</ScreenSubheader>
+      </ScreenStatusWrapper>
+      {success ? (
+        <ButtonStyled as={LinkStyled} fullWidth href={loginHref}>
+          {statusButtonText}
+        </ButtonStyled>
+      ) : (
+        <ButtonStyled as={LinkStyled} fullWidth onClick={onBackToStart}>
+          {statusButtonText}
+        </ButtonStyled>
+      )}
+    </ScreenContent>
+  );
+};
--- a/apps/web/lib/dictionary.tsx
+++ b/apps/web/lib/dictionary.tsx
@@ -89,6 +89,38 @@ const dictionary = {
         pl: "Zaloguj się",
       },
     },
+    successErrorScreen: {
+      status: {
+        error: {
+          subheader: {
+            en: "The account with this email already exists",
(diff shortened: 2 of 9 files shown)


apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
```diff
@@ -0,0 +1,91 @@
...
+  onBackToStart,
+}: SuccessErrorScreenProps) => {
+  const { dict, t } = useTranslate("SignUpPage");
+  const {
+    successErrorScreen: { status },
+  } = dict;
+
+  const statusHeaderText = useMemo(() => {
+    return success ? t(status.success.header) : t(status.error.header);
+  }, [status.error.header, status.success.header, success, t]);
```
aniasgit, line 68: Do we need useMemo?
```suggestion
  const statusHeaderText =  success ? t(status.success.header) : t(status.error.header);
```

apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
```diff
@@ -0,0 +1,91 @@
...
+import styled from "styled-components";
+import { Icon, LinkComponent } from "ui";
+
+const ScreenContent = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-direction: column;
+  height: 32.375rem;
+  font-family: "Signika", sans-serif;
```
aniasgit, line 14: I think you should use `font-family: "Signika", sans-serif;` on ScreenHeader and `font-family: inherit` on ScreenSubHeaderLink and remove font-family from here

apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { useMemo } from "react";
+import styled from "styled-components";
+import { Icon, LinkComponent } from "ui";
+
+const ScreenContent = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-direction: column;
+  height: 32.375rem;
```
aniasgit, line 13: This screen is too high on my device, but I think this will be corrected when we decide what are common content dimensions for all steps

Noshi96: We are still working on providing the same style on each screen, this will be changed in the future

aniasgit (approved):

apps/web/lib/dictionary.tsx
```diff
@@ -67,6 +67,30 @@ const dictionary = {
...
+            pl: "Powrót do rejestracji",
+          },
+          header: {
+            en: "The account with this email already exists",
+            pl: "Konto z tym adresem e-mail już istnieje",
+          },
+        },
+        success: {
+          link: {
+            en: "Navigate me to home",
```
monika-klokosinska-intive, line 84: @Noshi96 "Take me to the home page" sounds more natural.

packages/ui/Link/index.tsx
```diff
@@ -10,19 +11,20 @@ export const LinkComponent = ({
...
   ) : (
-    <LinkComponentAnchorStyled href={href}>
+    <LinkComponentAnchorStyled className={className} href={href}>
       {children}
     </LinkComponentAnchorStyled>
   );
 };
 
-export const LinkComponentAnchorStyled = styled.a<LinkComponentProps>`
+export const LinkComponentAnchorStyled = styled(Link)<LinkComponentProps>`
```
monika-klokosinska-intive, line 27: @Noshi96 Please define the styles above the main function - I know this might not be originally your code, but might as well fix this since you are already touching it.

apps/web/lib/dictionary.tsx
Noshi96: New texts from Figma added "Log in to your account"

packages/ui/Link/index.tsx
Noshi96: Okay done :D

apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
monika-klokosinska-intive: @Noshi96 I would avoid putting width and height in `rem` if possible. Hopefully, once the improved rwd nav is merged, the parent will sort out the width of the container. Instead of height, min-height should be a safer option as well.

apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
```diff
@@ -0,0 +1,116 @@
...
+  onBackToStart,
+}: SuccessErrorScreenProps) => {
+  const { push } = useRouter();
+  const { dict, t } = useTranslate("SignUpPage");
+  const {
+    successErrorScreen: { status },
+  } = dict;
+
+  // Probably a temporary solution
+  const navigateToHome = () => {
```
monika-klokosinska-intive, line 85: @Noshi96 I know you marked this as a temporary solution, but to make the component clearer either both `navigateToHome` and `onBackToStart` are coming from props or neither is. I would go with both of them coming from props - if flow controller provides all other steps transitions, let it handle this one as well.

Noshi96: @monika-klokosinska-intive I've changed it a bit. In our meeting on Thursday, we said that the button in the design should be a link, which makes sense when we have a positive scenario and the user data is valid. But otherwise, I think it's better to keep the button that supports the mentioned ```onBackToStart```. I made a conditioning renderer that renders the link in case of success and the button otherwise. The link gets the ```loginHref``` provided by the flow controller.

apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
Noshi96: @monika-klokosinska-intive I read an [article](https://medium.com/@sascha.wolff/dont-use-rem-em-for-paddings-margins-and-more-94e19026b000) that says rem/em should only be used for font-size. In our project, we have different approaches. Should we stick to fixed px for paddings, margins, width, height?

apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
monika-klokosinska-intive: @Noshi96 you are right, we will need the button in the negative scenario, to go back to step 1 and keep all the data (maybe except the incorrect e-mail)  the user already entered so they don't rage quit on us.

apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
monika-klokosinska-intive: @Noshi96 this is a great article! 
Width and height I feel like should stay away from rem/em if possible. To make width responsive we can use percentages, flex options (flex-basis, flex-grow etc) or grid options (fr units). `min-width`, `max-width` or `minmax`  are also there if we don't the element to go below / over a certain value.
Height is similar, but usually it needs to be more fluid, so it can look good with a small or a big amount of content. So height should be decided mostly be the content, again with `min-height` being available if you don't want it to go below a certain value. 
Of course there are always exceptions and there will be elements with a fixed width or height - for example a side nav that will be always 50px width or the main header bar that is 64px on desktop and 40px on mobile. 
Personally I have been sticking to fixed px margins and paddings when implementing designs for my current client and so far we haven't had any issues.

monika-klokosinska-intive:
@Noshi96 looking great except the mobile view. 
Zrzut ekranu z 2023-04-11 o 11 20 01

We will have to revisit this when putting every step together to make sure both elements' width are in sync. It just looks sloppy otherwise. @bdeluga please take note!

monika-klokosinska-intive (approved):

apps/web/app/(regflow)/sign-up/SuccessErrorScreen.tsx
Noshi96: @monika-klokosinska-intive Thank you for your insightful reply! :D

PR #37, opened 2023-04-10
feat: atomic component - datepicker #P23-214

AzulloCyll:
## Getting Started

Component is using third party library called [react-datepicker](https://www.npmjs.com/package/react-datepicker) so please run following commands to install needed packages.

```bash
npm install react-datepicker --save
```

```bash
npm install "@types/react-datepicker"
```

You can look at it in our storybook, or on demo page at route `/datepicker` [fastlink](https://patronage2023-js-ec5pu9sdp-wlechowicz.vercel.app/datepicker)

## Localization

Component is localized in two languages at now: English wich is default, and Polish. Language can be set using `lang="pl"` prop.

If you want to add more languages you need to import specific language locale from `date-fns/locale/...` like in example below

```bash
import es from 'date-fns/locale/es'
```

and register it in components datePickerLanguageConfig object.

## Known Issues

When using year dropdown it shows warning about [findDOMNode](https://stackoverflow.com/questions/71989683/finddomnode-is-deprecated-in-strictmode-using-react-datepicker-with-react-hook)

You can simply ignore it in console by right clicking that warning and choosing "Hide messages from react_devtools_backend.js". Of course this warning will not be visible in production version of our application.

Second issue is that we can't change years with keyboard.

## Usefull links

Demo:
[react-datepicker](https://reactdatepicker.com/)


packages/ui/DatePicker/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  onSelect,
+}: CustomDatePickerProps) => {
+  const [date, setDate] = useState<Date | null>(null);
+  const [locale, setLocale] = useState<Locale>();
+
+  useEffect(() => {
+    switch (lang) {
+      case "en":
+        setLocale(en);
+        break;
```
Noshi96, line 30: I think it should not be here. It should be in LanguageContext.tsx. DatePicker will only get locale/lang as props and set it in the <DatePicker>

Noshi96, line 30: I think you don't need either useState or switch. You can replace the switch with an object:
```javascript
type DatePickerLanguageConfigType = {
  [key: string]: Locale;
};

const datePickerLanguageConfig: DatePickerLanguageConfigType = {
  en: en,
  pl: pl,
  default: en,
};
```

Noshi96:
Good job on customizing the styles for the calendar. I like how it looks, also nice description of the PR :D

packages/ui/DatePicker/index.tsx
Noshi96, line 1: There is one problem with accessibility. Your year picker is actually wrapped in a div, making it inaccessible within the use of tabs and spaces. 

image

bdeluga:
@AzulloCyll Please resolve conflicts ! 👀 Except that the date picker looks very nice.  🚀

monika-klokosinska-intive (approved):

monika-klokosinska-intive (changes requested):
@AzulloCyll the deployments are failing - please take a look before merging!

AzulloCyll:
@monika-klokosinska-intive But on my computer it's building correctly I don'k know why it's not building.

monika-klokosinska-intive (approved):

wyola (approved):
Looks good to me!

zuzannapaw (approved):

PR #40, opened 2023-04-11
feat: profile screen component #P23-209

PawelPazderski:
### Avatar selection and name form for First name and Last name
ticket: https://tracker.intive.com/jira/browse/PATRO23-209

The component will need two props for future actions - `back() ` (go back one step) and `done()` where the values from the form and selected Avatar will be passed

### Changes
- introduces a number of changes to Avatar Selector appearance (eg. wrapper, mobile before and after avatar list shading elements)
- adding necessary fields to theme and dictionary files

*Due to onSubmitValidate operation problems inputs validation is being done with onChangeValidate*


apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
```diff
@@ -0,0 +1,177 @@
...
+        }}>
+        {({ submit }) => (
+          <StyledForm
+            onSubmit={(e) => {
+              e.preventDefault();
+              submit();
+            }}>
+            <H3Styled>{t(step3.title)}</H3Styled>
+            <SubtitleStyled>{t(step3.subtitle)}</SubtitleStyled>
+            <SeparatorLine />
```
wyola, line 83: There is a `Separator` component (its styles were updated in `PasswordScreen` since it displayed well in the storybook but not otherwise - the update is not yet merged to `main`)

apps/web/lib/dictionary.tsx
```diff
@@ -67,6 +67,44 @@ const dictionary = {
       en: "Hello from Sign-Up page",
       pl: "Witaj ze strony rejestracji",
     },
+    step3: {
```
wyola, line 70: I would name it rather "profileScreen" or similar - in case the order of steps changes or there are more steps. The name in the dictionary should be more general.

PawelPazderski: Yep you're right, I renamed the component during the process but didn't change that in dict, thanks!

apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
PawelPazderski: I added that for now I tried to use the Separator component though but didn't see anything somehow did't want to loose time on that I'll get back to it probably some detail to fix.

packages/ui/Input/index.tsx
```diff
@@ -32,6 +32,7 @@ export const Input = ({
   onChange,
   onFocus,
   onBlur,
+  onSubmit,
```
wyola, line 35: why do you add it here?

apps/web/lib/dictionary.tsx
```diff
@@ -67,6 +67,44 @@ const dictionary = {
...
+      },
+      lastNameInput:{
+        label: {
+          en: "Last name",
+          pl: "Nazwisko",
+        },
+        error : {
+          en: "The field cannot be empty",
+          pl: "Pole nie może być puste",
+        }
```
wyola, line 97: It is repeated both for first and last name - maybe it is better to move it one level up and name it "inputError"?

apps/web/lib/dictionary.tsx
```diff
@@ -67,6 +67,44 @@ const dictionary = {
...
+      },
+      backButton: { 
+        en: "Back", 
+        pl: "Powrót" 
+      },
+      submitButton: { 
+        en: "Create my account",
```
wyola, line 104: on design it is "Create account"

AzulloCyll:
image

Strange behavior. Maybe it need to switch display mode on other breakpoint. 
Overall component looks very good! :D Good job!

PawelPazderski:
> image
> 
> Strange behavior. Maybe it need to switch display mode on other breakpoint. Overall component looks very good! :D Good job!

I did't have that on my side - strange - must check what the problem is then, thanks!

apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
wyola: sure, we fixed it with Bartek K., you can see it on his branch later 🤗

apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
```diff
@@ -0,0 +1,177 @@
...
+  background-color: ${({ theme }) => theme.signUp.accent};
+  margin: 2em 0;
+`;
+
+type ProfileScreenProps = {
+  back: () => void;
+  done: () => void;
+} & React.HTMLProps<HTMLDivElement>;
+
+export const ProfileScreen = ({ back, done }: ProfileScreenProps) => {
```
bdeluga, line 60: Please consider also adding `profileInfo: {
    name: string;
    surname: string;
    avatar: string;
}` which will be passed from controller so user data persist in case of validation error e.g email taken.❌

apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
```diff
@@ -0,0 +1,177 @@
...
+const ButtonsWrapper = styled.div`
+  width: 100%;
+  margin-top: 1em;
+  display: flex;
+`;
+
+const CreateBtnWrapper = styled.div`
+  flex-grow: 1;
+  margin-left: 1em;
+`;
```
Noshi96, line 46: You can remove this wrapper and add gap:16px; in ButtonsWrapper. Same behavior less divs.

apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
monika-klokosinska-intive: @PawelPazderski @bdeluga is right, we need to pass the data from the user back to the parent component - so the parent component has access to them as well

PawelPazderski: Of course, wasn't sure at the moment of creating PR about the structure of the data for the Controller, I'll do as @bdeluga suggests then, as one object. Thanks

apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
PawelPazderski: Good idea, thanks

packages/ui/AvatarSelector/index.tsx
```diff
@@ -20,28 +22,49 @@ export const AvatarSelector = ({
...
         ))}
       </AvatarGridStyled>
-    </RelativeDiv>
+    </AvatarsWrapper>
   );
 };
 
+const AvatarsWrapper = styled.div`
+  position: relative;
+  width: 360px;
```
aniasgit, line 47: In the smallest screen (375px) this avatar selector is a bit too big. For the screen to look like in Figma, it should be 343px

packages/ui/theme.tsx
```diff
@@ -2,7 +2,7 @@
 import { ThemeProvider } from "styled-components";
 import { ReactNode } from "react";
 
-const colors = {
+export const colors = {
```
aniasgit, line 5: Do we need export this colors? I don't see any place where you import them.

PawelPazderski: it's no use here you're right

packages/ui/AvatarSelector/index.tsx
PawelPazderski: I would leave as much as we can here 348 px still fits good in the smallest width

apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
```diff
@@ -0,0 +1,169 @@
...
+              name="lastName"
+              initialValue={""}>
+              {({ value, setValue, onBlur, errors }) => (
+                <Input
+                  name="lastName"
+                  label={t(profileScreen.lastNameInputLabel)}
+                  value={value}
+                  onChange={(e) => setValue(e.currentTarget.value)}
+                  onBlur={onBlur}
+                  onInputCleared={() => setValue("")}
```
Alessiass, line 144:
```suggestion
                  onChange={(e) => setValue(e.currentTarget.value)}
                  onBlur={onBlur}
```
Clearing value also applies as a change. Is onInputCleared necessary?

apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
```diff
@@ -0,0 +1,169 @@
...
+  width: 100%;
+  height: 1px;
+  background-color: ${({ theme }) => theme.signUp.accent};
+  margin: 2em 0;
+`;
+
+type ProfileScreenProps = {
+  back: () => void;
+  done: ( profileInfo: {firstName: string; lastName: string; avatar: string} ) => void;
+} & React.HTMLProps<HTMLDivElement>;
```
bdeluga, line 54:
```suggestion 
type ProfileScreenProps = {
  back: () => void;
  done: ( profileInfo: {firstName: string; lastName: string; avatar: string} ) => void;
  profileData: {firstName:string,lastName:string,avatar:string};
} & React.HTMLProps<HTMLDivElement>;

bdeluga (changes requested):
Resolve suggestion, other than that, component works as expected ! 🚀

apps/web/app/(regflow)/sign-up/ProfileScreen.tsx
PawelPazderski: It's a input clearing option, it is provided by the input - is it neccessary? hard to tell

monika-klokosinska-intive:
@PawelPazderski please resolve conflicts!

bdeluga (approved):
Good job, also thanks for repairing AvatarSelector 🙌, approved. 🚀

Alessiass (approved):
Works like a charm

monika-klokosinska-intive (approved):

PR #46, opened 2023-04-17
refinement: P23-236-side-navbar-submenu

mpiwonski:
That is refinement of side navigation bar sub menu component (https://tracker.intive.com/jira/browse/PATRO23-236).


zuzannapaw (commented):
Component looks and works great. Although I would still work on SideNavBar placement to make it look like on design (there is still a gap on the right side when sidebar is closed). What do you also think about create dummy pages for users (when I click on some user in Teams, error 404 occurs)? Speaking of dummy pages, when I click on pages in Settings, the page name does not change in main card (it's always "Settings"). Also please check  `/budgets/[id]/page.tsx` and `budgets/page.tsx` and please delete unused imports. Well done for this component 🚀

mpiwonski:
> Component looks and works great. Although I would still work on SideNavBar placement to make it look like on design (there is still a gap on the right side when sidebar is closed). What do you also think about create dummy pages for users (when I click on some user in Teams, error 404 occurs)? Speaking of dummy pages, when I click on pages in Settings, the page name does not change in main card (it's always "Settings"). Also please check `/budgets/[id]/page.tsx` and `budgets/page.tsx` and please delete unused imports. Well done for this component 🚀

Thanks for your comments. Side navigation bar placement has been fixed and now it looks like it is shown in design. There won't be "Team" submenu as it has just been confirmed by clients (99%) - I was allowed to remove that item. As far as settings subpage is concerned, it has been designed to catch all possible instances of settings and as a result user is being redirected to one settings page as of now. As for budgets, that should be commented by @wlechowicz :)

packages/ui/SideNavigationBar/SubMenu/index.tsx
Noshi96, line 1: You can place style declarations above component declarations

Noshi96 (commented):
1. Opening a submenu expands the remaining content. We need to deal with this or consider overlapping submenus with the main card. 

image

2. The list of the main menu is a little too close to the left edge, and to the sub menu (design screen)

image

packages/ui/Input/SearchInput/index.tsx
```diff
@@ -1,18 +1,50 @@
...
+      <StyledInput
+        ref={inputRef}
+        placeholder={placeholder}
+        name={name}
+        id={id || name}
+        type={type}
+        value={value}
+        onChange={onChange}
+      />
+      <StyledIconSerch>
```
zuzannapaw, line 35: a little typo here in "StyledIconSerch" :)

packages/ui/SideNavigationBar/SubMenu/index.tsx
zuzannapaw: I agree with this comment:)

packages/ui/Input/SearchInput/index.tsx
```diff
@@ -1,18 +1,50 @@
-import styled from "styled-components";
-import { ReactNode } from "react";
+import styled, { ThemeContext } from "styled-components";
+import { Icon } from "../../Icon";
+import { StyledInputBase, StyledIcon } from "..";
+import React, { useContext, useRef } from "react";
```
zuzannapaw, line 4: I'm not sure if we still need `import React from "react"` 👀

zuzannapaw (approved):
Nicely applied changes. Sidenav placement looks much better 👍  Just a few little details from me, besides that, I think the component is overall fine

packages/ui/Input/SearchInput/index.tsx
monika-klokosinska-intive: @zuzannapaw @mpiwonski the ui package should work fine without importing `React` - please give it a try!

AzulloCyll: @mpiwonski I think that SearchInput shouldn't be nested in Input folder. SearchInput should go to `ui` directory

packages/ui/NavList/index.tsx
```diff
@@ -20,6 +20,7 @@ const NavListStyled = styled.ul`
   list-style: none;
   width: 100%;
   height: auto;
+  padding: 0;
```
bdeluga, line 23: No need for extra padding, its already declared in global.css 👀 

image

bdeluga (approved):
Minor tweaks, overall nice refinement 🥇

wlechowicz (changes requested):
The big change that needs to happen, is that the design shows 2 different behaviors of the side nav submenu - one where it pushes the content aside, squishing it, and the other, where it overlays the content.
We have to choose one behavior, and since we wouldn't be able to squish content - such as budget details - in a way that makes sense, we should choose to make the submenu show on top of the content, rather than squishing it to the side.

packages/ui/SideNavigationBar/SubMenu/index.tsx
mpiwonski: Thanks @Noshi96 and @zuzannapaw for your insights. The changes you have mentioned have been addressed. Please take a look at the latest commit.

packages/ui/Input/SearchInput/index.tsx
mpiwonski: Kudos @zuzannapaw eagle eye!

packages/ui/Input/SearchInput/index.tsx
mpiwonski: I thought it could lay under Input as it inherits some characteristics from it :)

packages/ui/Input/SearchInput/index.tsx
```diff
@@ -1,18 +1,50 @@
...
+        <Icon icon="search" color={theme.input.neutral} />
+      </StyledIconSerch>
+      {value && onInputCleared && (
+        <StyledIconCancel
+          onClick={(e) => {
+            e.preventDefault();
+            inputRef.current?.focus();
+            onInputCleared();
+          }}>
+          <Icon icon="cancel" color={theme.input.main} iconSize={20} />
```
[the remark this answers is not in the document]
mpiwonski: Thanks @bdeluga. Changes have been made accordingly.

packages/ui/NavList/index.tsx
mpiwonski: Yes, indeed, however it was for a purpose of storybook styling, but I deleted that anyway :)

mpiwonski:
> The big change that needs to happen, is that the design shows 2 different behaviors of the side nav submenu - one where it pushes the content aside, squishing it, and the other, where it overlays the content. We have to choose one behavior, and since we wouldn't be able to squish content - such as budget details - in a way that makes sense, we should choose to make the submenu show on top of the content, rather than squishing it to the side.

Thanks @wlechowicz for your insight. I have made a proper fix. Please take a look at the latest deployment.

packages/ui/SideNavigationBar/SideNavigationBarButton/index.tsx
```diff
@@ -1,43 +1,24 @@
...
 } & React.HTMLProps<HTMLDivElement>;
 
+type SpanProps = {
+  fontSize: number;
+} & React.HTMLProps<HTMLSpanElement>;
+
 const ListItemStyled = styled.li`
   list-style: none;
-  padding: 0;
+  /* padding: 0; */
```
monika-klokosinska-intive, line 21: @mpiwonski if you are not using this css value, delete it altogether instead of just commenting.
Leaving comments like this in code will eventually lead to bloat and mess (and you can always check previous history through git, so it's not like you are losing anything by deleting).

packages/ui/SideNavigationBar/SubMenu/index.tsx
```diff
@@ -25,56 +24,75 @@ type SubMenuProps = {
 } & React.HTMLProps<HTMLDivElement>;
 
 const SubMenuStyled = styled.div`
+  position: absolute;
+  top: 0;
+  left: 80px;
+
+  height: 100%;
+
```
monika-klokosinska-intive, line 32: @mpiwonski is there any reason for the new lines here? If not, delete them - a new line here and there contributes to the amount one has to scroll to get to the part of the code they are looking for.

mpiwonski: The sole reason was to separate common parts of CSS :)

wlechowicz (changes requested):
There's a bug - when I click Budgets, then, say, Bills, the menu disappears, which is OK. But when I click Budgets then, nothing happens. I have to click Budgets second time for it to kick in. I think it's interfering with your logic for closing the menu when the active item is clicked in the sidebar (not in the submenu).

mpiwonski:
> There's a bug - when I click Budgets, then, say, Bills, the menu disappears, which is OK. But when I click Budgets then, nothing happens. I have to click Budgets second time for it to kick in. I think it's interfering with your logic for closing the menu when the active item is clicked in the sidebar (not in the submenu).

Bug fixed.

AzulloCyll (approved):

Noshi96 (commented):
It looks very good after all the changes you made!🚀 I have one objection. When we use tab to navigate the navigation bar, when we click the space bar, the submenu opens, but the focus is still in the main navigation bar, so the next tab click does not move to the submenu. The behavior where when you click, the focus is redirected to the submenu can be achieved by placing the ```<SubMenu>``` in the ```<SideNavigationBarButton>``` as ```children```, and adding new conditional rendering from your ```activeFlag```

PawelPazderski (approved):
Ovarall it looks and works really nice, great job! I would suggest adding an option to close the menu by clicking outside it maybe, apart from that all looks good to me.

wlechowicz (approved):

PR #47, opened 2023-04-19
refactor: add link to logo #P23-251

wyola:
# Add link to the logo

Task: https://tracker.intive.com/jira/browse/PATRO23-251

- wrapper was changed from `div` to `a` and a `href` was added
- child element was changed to be an inline element like the parent ( from `div` to `span`)
- additionally, unnecessary exports were removed

### DIFF
--- a/packages/ui/Logo/index.tsx
+++ b/packages/ui/Logo/index.tsx
@@ -58,7 +58,7 @@ const LogoSignet = styled.span<LogoSignetProps>`
 
 export const Logo = ({ white, logoWidth = 144 }: LogoProps) => {
   return (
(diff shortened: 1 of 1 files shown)


wyola:
> Do we want to use it as an anchor tag instead of Link from next? In this case we are loosing whole aplication state.

wyola:
> Do we want to use it as an anchor tag instead of Link from next? In this case we are loosing whole aplication state.

oops, you are right

monika-klokosinska-intive (approved):

Noshi96 (approved):

Vectari (approved):

bdeluga (approved):
Approved🚀

PR #48, opened 2023-04-19
feat: create mock-data #P23-260

bdeluga:
[Chore](https://tracker.intive.com/jira/browse/PATRO23-260) was to create mocked data to use while backend is offline. `public/data.json` consist of few budgets.

### Structure :construction_worker:

- id: generated by `crypto.randomUUID()`
- date: `string` in YY/MM/DD format
- amount: either positive or negative `number` [linked with [chore](https://tracker.intive.com/jira/browse/PATRO23-248)]
- currency: `object` with ISO and locale [linked with [chore](https://tracker.intive.com/jira/browse/PATRO23-248)]
- category: `object` with id and name and icon with it's details [linked with [chore](https://tracker.intive.com/jira/browse/PATRO23-250)]
- description: `string`
- status: `string`
- creator: `string`

file `lib/types.ts` contains proposed type for transactions.

### Example usage :eyes:

`await fetch('/data.json').then(res=>res.json()).then(...) //do something with data`

### DIFF
--- a/apps/web/lib/types.ts
+++ b/apps/web/lib/types.ts
@@ -1,0 +1,21 @@
+export interface Transaction {
+  id: string;
+  date: string;
+  amount: number;
+  currency: {
+    ISO: string;
+    locale: string;
+  };
+  category: {
+    id: number;
+    name: string;
+    icon: {
+      name: string;
+      foreground: string;
+      background: string;
+    };
+  };
+  description: string;
+  status: string;
+  creator: string;
+}
--- a/apps/web/public/data.json
+++ b/apps/web/public/data.json
@@ -106,7 +106,7 @@
       "creator": "Mary Doe"
     },
     {
-      "id": "abe47732-673d-4bb1-a24b-c9a89b2dd5e5",
+      "id": "b8485583-4cc3-476a-a1ee-f25828cfc74c",
       "date": "2023-04-19",
       "amount": -47.99,
       "currency": {
--- a/apps/web/public/data.json
+++ b/apps/web/public/data.json
@@ -1,0 +1,214 @@
+{
+  "transactions": [
+    {
+      "id": "95dc1a67-cbe4-4e6b-8568-40a8e2f3d191",
+      "date": "2023-04-21",
+      "amount": -24.99,
+      "currency": {
+        "ISO": "USD",
+        "locale": "en-US"
+      },
+      "category": {
+        "id": 2,
+        "name": "Subscriptions",
+        "icon": {
+          "name": "play",
+          "foreground": "darkbrown",
+          "background": "brown"
+        }
+      },
+      "description": "Netflix",
+      "status": "Recurring",
+      "creator": "John Doe"
+    },
+    {
+      "id": "df8c7760-43a5-4379-a229-6495f3e67f9c",
+      "date": "2023-04-21",
+      "amount": -600,
+      "currency": {
+        "ISO": "USD",
+        "locale": "en-US"
+      },
+      "category": {
+        "id": 1,
+        "name": "Home spendings",
+        "icon": {
+          "name": "home",
+          "foreground": "darkgreen",
+          "background": "green"
+        }
+      },
+      "description": "Rent",
+      "status": "Done",
+      "creator": "Mary Doe"
+    },
+    {
+      "id": "4aa5af81-7745-42d8-bbd4-d804e407b8be",
+      "date": "2023-04-21",
+      "amount": 44.0,
+      "currency": {
+        "ISO": "EUR",
+        "locale": "de-DE"
+      },
+      "category": {
+        "id": 2,
+        "name": "Income",
+        "icon": {
+          "name": "money",
+          "foreground": "darkblue",
+          "background": "blue"
+        }
+      },
+      "description": "Lift share",
+      "status": "Done",
+      "creator": "Hans Niemann"
+    },
+    {
+      "id": "d26ed662-8807-40e2-b9c5-197862763151",
+      "date": "2023-04-20",
+      "amount": -9.99,
+      "currency": {
+        "ISO": "USD",
+        "locale": "en-US"
+      },
+      "category": {
+        "id": 2,
+        "name": "Subscriptions",
+        "icon": {
+          "name": "play",
+          "foreground": "darkbrown",
+          "background": "brown"
+        }
+      },
+      "description": "Spotify",
+      "status": "Cancelled",
+      "creator": "John Doe"
+    },
+    {
+      "id": "abe47732-673d-4bb1-a24b-c9a89b2dd5e5",
+      "date": "2023-04-20",
+      "amount": -77.99,
+      "currency": {
+        "ISO": "USD",
+        "locale": "en-US"
+      },
+      "category": {
+        "id": 4,
+        "name": "Grocery",
+        "icon": {
+          "name": "cart",
+          "foreground": "red",
+          "background": "none"
+        }
+      },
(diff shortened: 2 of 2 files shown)


apps/web/lib/types.ts
```diff
@@ -0,0 +1,21 @@
+export interface Transaction {
+  id: string;
+  date: string;
```
wlechowicz, line 3: This should probably be a timestamp (number), rather than a string. We will need it to group transaction per day and to format the timestamp into "Today", "Yesterday" and such (per design) by something like `Intl.RelativeTimeFormat` or a custom component.

apps/web/lib/types.ts
```diff
@@ -0,0 +1,21 @@
...
+    name: string;
+    icon: {
+      name: string;
+      foreground: string;
+      background: string;
+    };
+  };
+  description: string;
+  status: string;
+  creator: string;
```
wlechowicz, line 20: This is an object.

apps/web/public/data.json
```diff
@@ -0,0 +1,517 @@
...
+              "foreground": "#1E4C40",
+              "background": "#F1FBF6"
+            }
+          },
+          "description": "Rent",
+          "status": "Done",
+          "creator": {
+            "id": "5985e3b5-dbc5-46ba-bc71-781017fbc6f0",
+            "name": "Mary Doe",
+            "avatar": "8"
```
wlechowicz, line 29: This should be a URL, so "8.svg" or "8.png" for the placeholders, where for real users it would be a fully-qualified URL.

apps/web/public/data.json
```diff
@@ -0,0 +1,517 @@
...
+          }
+        },
+        {
+          "id": "4aa5af81-7745-42d8-bbd4-d804e407b8be",
+          "date": "2023-04-21",
+          "amount": 44.0,
+          "currency": {
+            "ISO": "EUR",
+            "locale": "de-DE"
+          },
```
wlechowicz, line 39: The shape of this should be the same as for the currency picker, so please negotiate this with @Vectari

wlechowicz (changes requested):
PR description doesn't match the interface, and the interface doesn't match the data 😅

apps/web/public/data.json
```diff
@@ -0,0 +1,517 @@
+{
+  "budgets": [
+    {
+      "id": "71ee9a04-6b27-423a-9caa-7d6a92335dae",
+      "name": "Main budget",
```
aniasgit, line 5: In budget we should also have start date, end date, limit, description (optional), icon and I think the currency is for whole budget not for every transaction individual.

bdeluga: Thanks, omw making changes 👀

apps/web/public/data.json
bdeluga: I'll change "ISO" to match "currency picker component", but locale is needed for "currency amount component" or do we get locale from current site's language ?

apps/web/public/data.json
AzulloCyll: I think id's of new objects should be generated on backend.

That's the type of object I've created so far.

`type budgetObjectType = {
  budgetName: string;
  budgetLimit: number;
  budgetDescription: string;
  budgetIcon: string;
  budgetDateStart: number;
  budgetDateEnd: number;
  budgetCurrency: "idon't know yet"
};`

so you can only change names to be more covenient.

bdeluga: Id will be generated on Backend, i just showed how could it looks like, as to names i would stick to what i already created. Nesting Budget prefix in Budget interface could make it confusing. 😵

AzulloCyll: I agree with you :) Just shown what I already have. I will change my atr names to that what agrreed with our whole team.

apps/web/public/data.json
```diff
@@ -0,0 +1,527 @@
...
+          "status": "Done",
+          "creator": {
+            "id": "5985e3b5-dbc5-46ba-bc71-781017fbc6f0",
+            "name": "Mary Doe",
+            "avatar": "8.svg"
+          }
+        },
+        {
+          "id": "e865e4d7-9760-49df-8010-648f8da8cb1c",
+          "date": "2023-03-11",
```
AzulloCyll, line 347: Some of dates are still strings insead of timestamps

AzulloCyll (commented):
Looks good to me! but please

apps/web/public/data.json
bdeluga: TYSM! for looking into this, and correcting me, I guess it's not my day 😓

monika-klokosinska-intive (approved):

AzulloCyll (approved):

apps/web/public/budget/36cc0777-2874-4d0c-a389-91280e45e836.json
```diff
@@ -0,0 +1,129 @@
...
+      "amount": -25.99,
+      "currency": {
+        "tag": "GBP",
+        "locale": "en-GB"
+      },
+      "category": {
+        "id": 4,
+        "name": "Grocery",
+        "icon": {
+          "name": "cart",
```
AzulloCyll, line 15: icon name should be `shopping_cart`

apps/web/public/budget/36cc0777-2874-4d0c-a389-91280e45e836.json
```diff
@@ -0,0 +1,129 @@
...
+    },
+    {
+      "id": "e865e4d7-9760-49df-8010-648f8da8cb1c",
+      "date": 1678492800000,
+      "amount": -50.27,
+      "currency": {
+        "tag": "GBP",
+        "locale": "en-GB"
+      },
+      "category": {
```
AzulloCyll, line 36: icon.name should be `directions_car`

apps/web/public/budget/36cc0777-2874-4d0c-a389-91280e45e836.json
```diff
@@ -0,0 +1,129 @@
...
+      "amount": -43.39,
+      "currency": {
+        "tag": "GBP",
+        "locale": "en-GB"
+      },
+      "category": {
+        "id": 3,
+        "name": "Car",
+        "icon": {
+          "name": "car",
```
AzulloCyll, line 65: car => directions_car

AzulloCyll (commented):
All icons `name = "car"` should be changed to `name="directions_car"`
all 'name = "cart" should be `name = "shopping_cart"`

apps/web/public/budget/36cc0777-2874-4d0c-a389-91280e45e836.json
```diff
@@ -0,0 +1,129 @@
...
+        "avatar": "8.svg"
+      }
+    },
+    {
+      "id": "df8c7760-43a5-4379-a229-6495f3e67f9c",
+      "date": 1677715200000,
+      "amount": -8996.3,
+      "currency": {
+        "tag": "GBP",
+        "locale": "en-GB"
```
AzulloCyll, line 109: Do we need to store locale in currency object? Maybe better we collect locale from our languageContext, or from loggedUser data?

Locale here is for showing currency amount, if we store it in budget.currency, there might be case that our budget's list shows currency at mixed styling which will be visually chaotic. I think better way is to show curencies formatted the same way according to current app language slelected by user. 

Same apllies to date formatting. 

@wlechowicz @monika-klokosinska-intive What do you think? Please correct me if i'm wrong.

PR #49, opened 2023-04-19
Feature/p23 212 selector

Vectari:
Hello! My task "Currency Select" - [PATRO23-212](https://tracker.intive.com/jira/browse/PATRO23-212)
You can test it here: https://patronage2023-js-kjzfabmx5-wlechowicz.vercel.app/ then click on Budgets > Add New Budget.

I use Radix Select so remember about `npm i` :)


packages/ui/CurrencySelect/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+      </SelectTrigger>
+      {supportingLabel && (
+        <StyledSupportingLabel hasError={hasError}>
+          Choose Currency
+        </StyledSupportingLabel>
+      )}
+      <SelectPortal>
+        <SelectContent position="popper">
+          <Select.Viewport>
+            {currency.map((currency) => (
```
monika-klokosinska-intive, line 169: @Vectari when mapping a unique `key` is necessary on the mapped child element.

monika-klokosinska-intive:
@Vectari any chance you could style the storybook example to look more like the actual usage in `Create New Budget`? The storybook version of the component has ugly borders for the active element and a lot of unnecessary padding for each option. The version used in the modal looks fine.

packages/ui/CurrencySelect/index.tsx
AzulloCyll: key could be a curency.tag

packages/ui/CurrencySelect/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+              <SelectItem value={currency.id}>
```
AzulloCyll, line 170: Since we using currency tags in our shema sended to backend, we can use tags for value here.
`value={currency.id}` => `alue={currency.tag}>`

Moreover we can get rid of id's from currency array.

PR #53, opened 2023-04-22
feat: budget icon component #P23-245

mpiwonski:
Budget Icon component: https://tracker.intive.com/jira/browse/PATRO23-245.

- Component receives `icon` prop that determines which icon is displayed
- Component size depends on the size of a parent (`width: 100%; height: 100%`)
- Component is also used in `IconPicker` component

There's an example of component usage in storybook that you can take a look at. It's draft, so there may be some differences compared with Figma design.

**Most important question: is it enough to set width/height to 100% as to achieve dependency on parent element?**

### DIFF
--- a/packages/ui/BudgetIcon/index.tsx
+++ b/packages/ui/BudgetIcon/index.tsx
@@ -1,0 +1,35 @@
+import styled from "styled-components";
+import { ReactNode } from "react";
+import { Icon, IconType } from "../Icon";
+
+type BudgetIconProps = {
+  icon?: IconType;
+  children: ReactNode;
+};
+
+const BudgetIconStyled = styled.div`
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  height: 100%;
+  width: 100%;
+  background-color: ${({ theme }) => theme.budgetIcon.background};
+  color: ${({ theme }) => theme.budgetIcon.main};
+`;
+
+export const BudgetIcon = ({ icon, children }: BudgetIconProps) => {
+  return (
+    <BudgetIconStyled>
+      {icon ? (
+        <>
+          <Icon icon={icon} iconSize={40} />
+          {children}
+        </>
+      ) : (
+        <>{children}</>
+      )}
+    </BudgetIconStyled>
+  );
+};
--- a/packages/ui/IconPicker/index.tsx
+++ b/packages/ui/IconPicker/index.tsx
@@ -2,11 +2,11 @@
 import { ReactNode, useEffect, useState } from "react";
 import {
   IconPickerStyled,
-  IconAndButtonWrapperStyled,
   IconsSelectorStyled,
   EditButtonStyled,
   SelectIconButtonStyled,
 } from "./iconPicker.styled";
+import { BudgetIcon } from "../BudgetIcon";
 import { Icon, IconType } from "../Icon";
 
 type IconPickerProps = {
@@ -90,16 +90,11 @@ export const IconPicker = ({
 
   return (
     <IconPickerStyled>
-      <IconAndButtonWrapperStyled>
-        {currentIcon ? (
-          <Icon icon={currentIcon} iconSize={40} />
-        ) : (
-          <>{children}</>
-        )}
+      <BudgetIcon icon={currentIcon}>
         <EditButtonStyled onClick={handleEditButtonClick}>
           <Icon icon="edit" iconSize={12} />
         </EditButtonStyled>
-      </IconAndButtonWrapperStyled>
+      </BudgetIcon>
       {iconSelectorVisible && (
         <IconSelector
           icons={icons}
--- a/apps/storybook/stories/BudgetIcon.stories.tsx
+++ b/apps/storybook/stories/BudgetIcon.stories.tsx
@@ -1,0 +1,69 @@
+import React from "react";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import styled from "styled-components";
+
+import { BudgetIcon } from "ui";
+
+export default {
+  title: "BudgetIcon",
+  component: BudgetIcon,
+} as ComponentMeta<typeof BudgetIcon>;
+
+const Wrapper = styled.div`
+  display: flex;
+  gap: 10px;
+  min-height: 80px;
+`;
+
+const BudgetIconWrapper = styled.div`
+  display: flex;
+  min-width: 80px;
+`;
+
+const Title = styled.span`
+  font-size: 24px;
+  font-weight: 600;
+  color: ${({ theme }) => theme.budgetIcon.main};
+`;
+
+const Description = styled.span`
+  font-size: 12px;
+  color: "#626262";
+`;
+
+const Info = styled.div`
+  display: flex;
+  flex-grow: 1;
+`;
+const TitleDescriptionWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  align-items: flex-start;
+  gap: 20px;
+`;
+
+const BudgetIconComponent: ComponentStory<typeof BudgetIcon> = ({
+  ...args
+}) => {
+  return (
+    <Wrapper>
+      <BudgetIconWrapper>
+        <BudgetIcon {...args} />
+      </BudgetIconWrapper>
+      <Info>
+        <TitleDescriptionWrapper>
+          <Title>Bills</Title>
+          <Description>This is a description of Bills budget.</Description>
+        </TitleDescriptionWrapper>
+      </Info>
+    </Wrapper>
+  );
+};
+
+export const BudgetIconWithBudgetNameAndDescription = BudgetIconComponent.bind(
+  {}
+);
+BudgetIconWithBudgetNameAndDescription.args = {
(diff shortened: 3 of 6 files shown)


packages/ui/BudgetIcon/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const BudgetIcon = ({ icon, children }: BudgetIconProps) => {
+  return (
+    <BudgetIconStyled>
+      {icon ? (
+        <>
+          <Icon icon={icon} iconSize={40} />
+          {children}
+        </>
+      ) : (
+        <>{children}</>
```
AzulloCyll, line 31: It's good to show any Icon instead of showing nothing. Consider adding some placeholder like icon with questionmark, or any other suitable icon in this case.

AzulloCyll (approved):
Only one change and it will be good to me. 
Good idea on reafactoring IconPicker to use your component:)

packages/ui/BudgetIcon/index.tsx
zuzannapaw, line 31: Could you please explain to me why if an icon wasn't passed ,`children` are shown? Am I thinking correctly that `children`  in this case can be only a button that modifies the icon (shown in icon picker)?

mpiwonski: That's a good point, thank you! ☺️

zuzannapaw (approved):
I red that this component is a draft so it can look diffrently from figma design, but my small suggestions would be to add higher value of `iconSize` and add margin-right to the component to make wider space between component and budget's name. Otherwise it looks good to me 👍

Alessiass (approved):
Looks good to me 🎉

packages/ui/BudgetIcon/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const BudgetIconStyled = styled.div`
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  height: 100%;
+  width: 100%;
```
wyola, line 17: If you have `height` and `width` set to 100%, you have to then set the size on the parent. It would be better to pass a `className` as a prop, so it is possible to add extra styles when using the component.

With current approach, you always need to wrap `BudgetIcon` with another element just to add the dimensions.

monika-klokosinska-intive: @mpiwonski @wyola the icon taking the parent's size should be the default behaviour. That being said passing `className` is not a bad idea as we might need to overwrite the default styles in some specific case in the future anyway - better that the component is ready for that.

packages/ui/BudgetIcon/index.tsx
monika-klokosinska-intive: I believe this is to accommodate the case where:
- Budget Icon is rendered inside the IconPicker
- we don't want to pick any "default" icon in the IconPicker, instead it remains empty so the user knows they haven't selected anything yet.

aniasgit: In the previous version, if there was no icon, children were used to provide an element, e.g. with a text about the need to select an icon. I thought a blank circle wouldn't look good in this case and it would be nice to be able to give some information there. Look in old IconPicker storybook:
https://patronage2023-js-storybook-git-feature-p23-21-074fff-wlechowicz.vercel.app/?path=/story/icon-picker--without-default-icon

packages/ui/IconPicker/index.tsx
```diff
@@ -90,16 +90,11 @@ export const IconPicker = ({
 
   return (
     <IconPickerStyled>
-      <IconAndButtonWrapperStyled>
-        {currentIcon ? (
-          <Icon icon={currentIcon} iconSize={40} />
-        ) : (
-          <>{children}</>
-        )}
+      <BudgetIcon icon={currentIcon}>
```
aniasgit, line 93: There was a possibility to provide for example text to icon picker, when there wasn't default icon. You removed this.

packages/ui/BudgetIcon/index.tsx
Noshi96, line 31: It can look simple like this:

```javascript
export const BudgetIcon = ({ icon, children }: BudgetIconProps) => {
  return (
    <BudgetIconStyled>
      {icon ? <StyledIcon icon={icon} /> : children}
    </BudgetIconStyled>
  );
};
```

packages/ui/BudgetIcon/index.tsx
```diff
@@ -0,0 +1,35 @@
+import styled from "styled-components";
+import { ReactNode } from "react";
+import { Icon, IconType } from "../Icon";
+
+type BudgetIconProps = {
+  icon?: IconType;
+  children: ReactNode;
+};
```
Noshi96, line 8:
```suggestion
type BudgetIconProps = {
  icon?: IconType;
} & React.HTMLProps<HTMLDivElement>;
```
TS will not shout at this 😄

packages/ui/BudgetIcon/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  background-color: ${({ theme }) => theme.budgetIcon.background};
+  color: ${({ theme }) => theme.budgetIcon.main};
+`;
+
+export const BudgetIcon = ({ icon, children }: BudgetIconProps) => {
+  return (
+    <BudgetIconStyled>
+      {icon ? (
+        <>
+          <Icon icon={icon} iconSize={40} />
```
Noshi96, line 27:
```suggestion
          <Icon icon={icon} />
```
We don't want this if we want to get responsive icon

packages/ui/IconPicker/index.tsx
```diff
@@ -90,16 +90,11 @@ export const IconPicker = ({
 
   return (
     <IconPickerStyled>
-      <IconAndButtonWrapperStyled>
-        {currentIcon ? (
-          <Icon icon={currentIcon} iconSize={40} />
-        ) : (
-          <>{children}</>
-        )}
```
Noshi96, line 98: You can restore the almost old version and Icon Picker will work as before 😃 

```javascript
    <IconPickerStyled>
      {currentIcon ? (
        <BudgetIcon icon={currentIcon} />
      ) : (
        <BudgetIcon>{children}</BudgetIcon>
      )}
      <EditButtonStyled onClick={handleEditButtonClick}>
        <Icon icon="edit" iconSize={12} />
      </EditButtonStyled>
```

Noshi96 (changes requested):
In my opinion, if we want a responsive icon size, we should remove the fixed value in the icon size and manipulate the font-size.

I am also skeptical about passing children to ```BudgetIcon```. The responsibility of ```BudgetIcon``` should be to display only the icon on a rounded background. Your example in the storybook shows the budget header, where we should have separate elements 1. ```BudgetIcon``` and 2. ```div``` with budget title and description, my point is that ```BudgetIcon``` is not responsible for showing us the budget header. For me, ```BudgetIcon``` could get an optional ReacNode prop instead of ```children```, but below I show the solution with ```children``` in terms of getting compatibility with ```IconPicker``` (which would also be possible with an optional ```ReacNode``` instead of ```children```). For compatibility with ```IconPicker```, I show the following approach, where we do not pass an editable button to ```BudgetIcon```, but only optional ReactNode ```children```.

packages/ui/BudgetIcon/index.tsx
mpiwonski: Fixed.

packages/ui/BudgetIcon/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+type BudgetIconProps = {
+  icon?: IconType;
+  children: ReactNode;
+};
+
+const BudgetIconStyled = styled.div`
+  position: relative;
+  display: flex;
+  justify-content: center;
```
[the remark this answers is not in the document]
mpiwonski: As discussed on the call, we won't go into that direction, but thanks for your insight :)

packages/ui/BudgetIcon/index.tsx
mpiwonski: As discussed on the call, width and height will stay as it is, but I have added className prop. Thanks for your insights :)

packages/ui/BudgetIcon/index.tsx
mpiwonski: As discussed on the call, icon component will now be extended and it will set font size in em units.

packages/ui/BudgetIcon/index.tsx
mpiwonski: @aniasgit is right. There will be text if no icon has been selected.

mpiwonski: The old version has been brought back.

packages/ui/IconPicker/index.tsx
mpiwonski: Its back :)

mpiwonski (commented):
Thanks for all the comments. Changes have been made. Please take a look at another version!

aniasgit (approved):
Looks good to me 👍

monika-klokosinska-intive (approved):

Noshi96 (approved):

PR #54, opened 2023-04-23
feat: three dots component #P23-249

zuzannapaw:
**Info:**

- link to task in jira: https://tracker.intive.com/jira/browse/PATRO23-249 👀,
- ThreeDots component based on `DropdownMenu` from radix (https://www.radix-ui.com/docs/primitives/components/dropdown-menu),
- component accepts `items` as prop in order to be more generic - it can be used in diffrent places with diffrent items,
- additional design: background on hover, outline on focus, border-radius on items.

**Main question:**

Does `items` array should be an array of strings, links or buttons? - (key aspect of component's functionality - will it relocate user to another page?) 🤔  **update ->** **problem fixed** ✔️ 

💥 **Current problem:** 

Storybook fails when I use `useTranslate` hook inside component - figuring out why.  **update ->** **problem fixed** ✔️


packages/ui/ThreeDots/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <DropdownMenu.Root>
+      <DropdownMenuTriggerStyled asChild>
+        <button>
+          <Icon icon="more_vert" />
+        </button>
+      </DropdownMenuTriggerStyled>
+      <DropdownMenu.Portal>
+        <DropdownMenuContentStyled side="right">
+          {items.map((item, i) => {
```
AzulloCyll, line 46: `i` variable is not used in further code.

packages/ui/ThreeDots/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+    background: ${({ theme }) => theme.threeDots.backgroundHover};
+  }
+`;
+
+export const ThreeDots = ({ items }: ThreeDotsProps) => {
+  return (
+    <DropdownMenu.Root>
+      <DropdownMenuTriggerStyled asChild>
+        <button>
+          <Icon icon="more_vert" />
```
AzulloCyll, line 41: You need to add color to icon. In our project it's not black.

AzulloCyll (approved):
Please correct two little errors, and it's goot to merge.

packages/ui/ThreeDots/index.tsx
zuzannapaw: thank you, I will fix that :)

packages/ui/ThreeDots/index.tsx
zuzannapaw: good eye! On my way to fix it 👀

packages/ui/ThreeDots/index.tsx
```diff
@@ -0,0 +1,53 @@
+"use client";
+import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
+import { Icon } from "ui";
+import styled from "styled-components";
+
+type ThreeDotsProps = {
+  items: Array<string>;
+} & React.HTMLProps<HTMLSelectElement>;
```
Alessiass, line 8: We should somehow pass the function for each of the items. Just string isn't enough.

Alessiass (changes requested):

packages/ui/ThreeDots/index.tsx
zuzannapaw: @Alessiass  thank you for that comment. Why do you think of a function? Because if each of items will relocate user to another page I think that array of links would be accurate? I don't know the main idea of that so that's why i put this `Main question` in PR's comment 😄

packages/ui/ThreeDots/index.tsx
zuzannapaw: @AzulloCyll done :)

packages/ui/ThreeDots/index.tsx
zuzannapaw: @AzulloCyll done :)

packages/ui/ThreeDots/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+const DropdownMenuContentStyled = styled(DropdownMenu.Content)`
+  margin-left: 8px;
+`;
+
+const DropdownMenuItemStyled = styled(DropdownMenu.Item)`
+  width: 100%;
+  padding: 8px;
+  font-size: 14px;
+  border-radius: 5px;
+  cursor: pointer;
```
wyola, line 26: why not throw in some nice styles here to spice it up a bit 🚀 I know there is no design, but you could make it look like this one:

image

```suggestion
const DropdownMenuContentStyled = styled(DropdownMenu.Content)`
  margin-left: 8px;
  border: 1px solid #E1E1E1;
  box-shadow: 0px 2px 8px rgba(32, 37, 50, 0.08), 0px 2px 4px rgba(32, 37, 50, 0.03);
  border-radius: 16px;
  background-color: white;
  overflow: hidden;
`;

const DropdownMenuItemStyled = styled(DropdownMenu.Item)`
  padding: 8px;
  font-size: 14px;
  border-radius: 5px;
  cursor: pointer;
```
this way you arrive at this:

image

nevertheless, you need to add at least `background-color: white;` on `DropdownMenuContentStyled` because right now it is transparent and upon opening it would not look good if there is anything under it 🤗

PS colors should come from the theme, I forgot about it here 🤭

packages/ui/ThreeDots/index.tsx
wyola, line 8: only `items` prop is used. I think that `React.HTMLProps<HTMLSelectElement>` can be removed

```suggestion
};
```

packages/ui/ThreeDots/index.tsx
```diff
@@ -0,0 +1,53 @@
+"use client";
+import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
+import { Icon } from "ui";
+import styled from "styled-components";
+
+type ThreeDotsProps = {
+  items: Array<string>;
+} & React.HTMLProps<HTMLSelectElement>;
+
+const DropdownMenuTriggerStyled = styled(DropdownMenu.Trigger)`
```
wyola, line 10: Right now, when the element is focused with keyboard there's no visual indication that it's active. Adding styles for `:focus` state (maybe an outline?) should fix it.

packages/ui/ThreeDots/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+    background: ${({ theme }) => theme.threeDots.backgroundHover};
+  }
+`;
+
+export const ThreeDots = ({ items }: ThreeDotsProps) => {
+  return (
+    <DropdownMenu.Root>
+      <DropdownMenuTriggerStyled asChild>
+        <button>
+          <Icon icon="more_vert" color="#7E7E7E"/>
```
wyola, line 41: Use color from the theme

packages/ui/ThreeDots/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+        </button>
+      </DropdownMenuTriggerStyled>
+      <DropdownMenu.Portal>
+        <DropdownMenuContentStyled side="right">
```
wyola, line 45: maybe consider passing the position as a prop? I am thinking especially about mobile, when there probably won't be enough space on the right

wyola (commented):
Looking good! on general note, I would change component's name simply to "DropdownMenu" or something more generic than "ThreeDots"? 🤔

packages/ui/ThreeDots/index.tsx
Alessiass: Just any function should do the job. It's not this dropdown menu's concern what the button will do. If we want to redirect to other path we will add something that will replace pathname, if we want to remove budget we will add fetch function and so on.

monika-klokosinska-intive:
@zuzannapaw since we don't have any idea what the menu under the the three dots will actually do 🙃 how about instead of passing `items` we will just render `children` as the dropdown submenu? This way we should be prepared for everything that might come our way - buttons, links, custom components etc,

packages/ui/ThreeDots/index.tsx
aniasgit: I agree with Bartek

packages/ui/ThreeDots/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  &:hover {
+    background: ${({ theme }) => theme.threeDots.backgroundHover};
+  }
+`;
+
+export const ThreeDots = ({ items }: ThreeDotsProps) => {
+  return (
+    <DropdownMenu.Root>
+      <DropdownMenuTriggerStyled asChild>
+        <button>
```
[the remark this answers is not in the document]
zuzannapaw: Thank you @wyola for your comments and ideas with design! ✨  On my way to apply your suggestions!

packages/ui/ThreeDots/index.tsx
zuzannapaw: @Alessiass @aniasgit I fixed this problem, now dropdown renders children inside itself. Currently children are created in `TransactionDropdownMenu.stories.tsx` and it's an array full of components. We can use any components that we like ☺️

packages/ui/ThreeDots/index.tsx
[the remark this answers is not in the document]
zuzannapaw: @wyola I have applied your suggestions - with code, design and component's name.

aniasgit:
> Storybook fails when I use `useTranslate` hook inside component - figuring out why.

useTranslate and dictionary are parts of web applications, they works with language contex, which is only in web application, not in storybook.

packages/ui/TransactionDropdownMenu/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+const DropdownMenuTriggerStyled = styled(DropdownMenu.Trigger)`
+  background-color: transparent;
+  border: 0;
+  outline: 0;
+  cursor: pointer;
+
+  &:focus {
+    outline: ${({ theme }) => theme.transactionDropdownMenu.outlineFocus};
+  }
+`;
```
wyola, line 21: fix for focus outline on tab (you might adjust the thickness value though, I did not look into that)
```suggestion
const DropdownMenuTriggerStyled = styled(DropdownMenu.Trigger)`
  background-color: transparent;
  border: 0;
  cursor: pointer;

  &:focus {
    outline: ${({ theme }) => theme.transactionDropdownMenu.outlineFocus} 2px solid;
  }
`;
```

zuzannapaw: @wyola great, thank you, I have applied your fix!

Alessiass (approved):

packages/ui/TransactionDropdownMenu/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  border-radius: 5px;
+  cursor: pointer;
+
+  &:focus {
+    outline: ${({ theme }) => theme.transactionDropdownMenu.outlineFocus};
+  }
+  &:hover {
+    background: ${({ theme }) => theme.transactionDropdownMenu.backgroundHover};
+  }
+`;
```
wyola, line 51: if you change those styles like below:

```suggestion
export const DropdownMenuItemStyled = styled(DropdownMenu.Item)`
  padding: 6px 11px;
  font-size: 14px;
  cursor: pointer;  
  border: transparent 2px solid; 

  &:first-child {
    border-top-left-radius: 16px;
    border-top-right-radius: 16px;
  }

  &:last-child {
    border-bottom-left-radius: 16px;
    border-bottom-right-radius: 16px;
  }

  &:focus {
    border-color: ${({ theme }) => theme.transactionDropdownMenu.outlineFocus};    
  }

  &:hover {
    background: ${({ theme }) => theme.transactionDropdownMenu.backgroundHover};
    outline: 0;
  }
`;
```

there will be a nicer border effect on items when accessing the menu from keyboard:

https://user-images.githubusercontent.com/80450315/234214239-5a703bfb-8bd2-4cd5-a11b-8c5177154e5f.mp4

packages/ui/TransactionDropdownMenu/index.tsx
bdeluga, line 1: I think the way you showcase component might be misleading to use. I don't think DropMenu should export its Item for other components to use. I was thinking maybe It should also expect us to recieve array of components to render.

image

Then just map through components and render them in this component instead of just passing `children`

image

This would be much easier for us to use - just passing items as object array instead of having to do all this array parkour. 

image


### Usage 🚀 

image

 ###  ❤️ Nontheless your solution isn't wrong, we could stick to it ❤️

packages/ui/TransactionDropdownMenu/index.tsx
monika-klokosinska-intive: @zuzannapaw @wyola has some 💯 ideas here. We definitely need to add outlines for keyboard access, otherwise it looks un-accessible even though it is.

bdeluga (changes requested):
Tell me what you think 🤔 👀

packages/ui/TransactionDropdownMenu/index.tsx
zuzannapaw: @bdeluga  thank you for your comments! Indeed that's a good solution - I have thought about it. But as @monika-klokosinska-intive  said, for now the dropdownmenu could accept `children` to be prepared for anything that will be rendered inside. Maybe in future, when we will know more details, it will be a good idea to make a component's refinement 👀 
image

monika-klokosinska-intive: @zuzannapaw I was about to leave the same comment as @bdeluga :D. It's much better to just pass a `components` or `items` array without `DropDownMenuItemStyled`. 

`DropDownMenuItemStyled` should be added inside the map function. As a user of this component, I just want my array of stuff to be rendered inside the dropdown. I shouldn't have to know the inner workings of the drop down  (wrapping each item in `DropDownMenuItemStyled`, importing `DropDownMenuItemStyled`) to get that done.

Since `DropDownMenuItemStyled` has to be added to every item anyway, it makes it much simpler to do it inside the map. It will also make the actual usage of the component inside our app simpler - we won't have to add `DropDownMenuItemStyled` multiple times inside the `items` prop.

monika-klokosinska-intive (changes requested):
@zuzannapaw Please take a look at the excellent suggestions by @bdeluga @wyola! Once they are applied and the conflicts are merged, we are good to go 👍

bdeluga (approved):
Approving ahead of time 🚀

packages/ui/TransactionDropdownMenu/index.tsx
zuzannapaw: @bdeluga @monika-klokosinska-intive  thank you for your comments, I will fix it in a moment!

zuzannapaw: @monika-klokosinska-intive @bdeluga  I have resolved conflicts and changed component as desired! Is it ready to merge?

zuzannapaw: Although I can still see some problems with deployment :(

packages/ui/TransactionDropdownMenu/index.tsx
wyola: @zuzannapaw it's working great! sorry for being a pain! 🤗

packages/ui/TransactionDropdownMenu/index.tsx
wyola: probably because of the limits on the vercel :( 

image

monika-klokosinska-intive (approved):

PR #55, opened 2023-04-23
feat: info tile for basic budget information #P23-246

PawelPazderski:
## The Info Tile component for the budget basic information

Ticket for the task: https://tracker.intive.com/jira/browse/PATRO23-246

The component receives two type of props:

- `label` - which is a string
- `dataToRender` - representing a ReactNode

### Example of usage:

Data:

```javascript
const exampleData = {
  dataToRender: (
    <>
      <Icon icon="subscriptions"/>
      <span>Awesome Icon</span>
    </>
  ),
  label: "Some Icon",
};
```

Code:
```javascript
<InfoTile label={exampleData.label} dataToRender={exampleData.dataToRender} />
```

Result: 

image

---
To see the live example of tiles rendered from dummy_data on this branch go to sign-up/playground
or see the link below:
https://patronage2023-js-web-git-feature-p23-246-info-tile-wlechowicz.vercel.app/sign-up/playground


packages/ui/InfoTile/infoTile.styled.tsx
```diff
@@ -0,0 +1,37 @@
+import styled from "styled-components";
+import { device } from "../../../apps/web/lib/media-queries";
+
+export const InfoTileStyled = styled.div`
+  border: 1px solid ${({ theme }) => theme.infoTile.border};
+  color: ${({ theme }) => theme.infoTile.label};
+  padding: 16px 32px 16px 16px;
```
wyola, line 7: Correct the padding - top and bottom should be 8px

packages/ui/InfoTile/infoTile.styled.tsx
```diff
@@ -0,0 +1,37 @@
...
+  flex-direction: column;
+`;
+
+export const InfoValueWrapper = styled.div`
+  font-family: "Signika", sans-serif;
+  display: flex;
+  gap: 5px;
+  color: ${({ theme }) => theme.infoTile.value};
+  font-size: 16px;
+  font-weight: 600;
```
wyola, line 21: add `margin-top: 4px;`

image

packages/ui/InfoTile/infoTile.styled.tsx
```diff
@@ -0,0 +1,37 @@
...
+  font-size: 12px;
+  line-height: 24px;
+  display: flex;
+  flex-direction: column;
+`;
+
+export const InfoValueWrapper = styled.div`
+  font-family: "Signika", sans-serif;
+  display: flex;
+  gap: 5px;
```
wyola, line 18: `gap` should be 8px I think 

image

packages/ui/InfoTile/index.tsx
```diff
@@ -0,0 +1,18 @@
+"use client"
+
+import {InfoTileStyled, InfoValueWrapper} from './infoTile.styled'
+
+export type InfoTileProps = {
+  label: string;
+  children: string | JSX.Element;
```
wyola, line 7: you could change it like below, this way it includes both string and components (basically anything what React can render) -> https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/basic_type_example/#useful-react-prop-type-examples

```suggestion
  children: React.ReactNode;
```

wyola (commented):
Looking good, just some minor styling issues. I am not sure about moving styles to a separate file - there is so little code that I don't know if it's necessary (unless we established that we should do it always, and I missed it 🫥)

packages/ui/InfoTile/index.tsx
monika-klokosinska-intive: @PawelPazderski instead of rendering `children`, try passing a component to render as a prop named `value` as the story specified (though admittedly the Jira story could use some polish). Zuza's `NavItem` inside `NavList` accepts `componentToRender` as a prop - this is the solution we are looking for here. 
I know it doesn't make that much difference in the end (component will work and look the same), but it's a good learning experience regardless.

monika-klokosinska-intive:
@PawelPazderski storybook deployment crashed - please look into it as it is a blocker for merging.

packages/ui/InfoTile/infoTile.styled.tsx
PawelPazderski: @wyola thanks a lot for all the suggestions, I'll do some style corrections taking them into mind

packages/ui/InfoTile/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import { ReactHTMLElement } from 'react';
+import {InfoTileStyled, InfoValueWrapper} from './infoTile.styled'
+
+export type InfoTileProps = {
+  label: string;
+  children: React.ReactNode
+} 
+
+export const InfoTile = ({ label, children }: InfoTileProps) => {
```
monika-klokosinska-intive, line 11: @PawelPazderski I apologise for not being clear enough. The only part of NavList I wanted lifted was the `dataToRender` part. So the component should be:
```
export const InfoTile = ({ label, dataToRender }: InfoTileProps) => {

  return (
    <InfoTileStyled>
      {label}
      <InfoValueWrapper>{dataToRender}</InfoValueWrapper>
    </InfoTileStyled>
  );
};
```

And the example props passed to InfoTile should look like this:

```
InfoTileDemo.args = {
  dataToRender: (
    <>
      <div style={{fontSize: "24px"}}>
        <Avatar src="/avatar.svg" />
      </div>
      
    </>
  ),
  label: "Some avatar",
}
```
We don't require a list of InfoTiles, so the `InfoTileList` is unnecessary. All we want here is a component `InfoTile` with props of `label` and `dataToRender`.

I am sorry for the extra work you did, I hope it was at least of educational value :).

monika-klokosinska-intive (changes requested):
@PawelPazderski please take a look at my newest comment

packages/ui/InfoTile/index.tsx
PawelPazderski: Right no worries, I'll change that

PawelPazderski: OK, I guess the changes are done according to your suggestions, only problem is the fonts in storybook seem not to work when passing a complex structure - works fine with a string, in other cases the inheritance doesn't seem to work

packages/ui/InfoTile/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  padding: 8px 32px 8px 16px;
+  border-radius: 8px;
+  font-size: 12px;
+  line-height: 24px;
+  display: flex;
+  flex-direction: column;
+`;
+
+export const InfoValueWrapper = styled.div`
+margin-top: 4px;
```
aniasgit, line 17: lack of spaces ;)

PawelPazderski: thanks

bdeluga (approved):
Nice 🚀

bdeluga:
Got some conflicts to resolve 👀

PR #58, opened 2023-04-25
feat: (Next-)Auth.js credentials log-in intergration #P23-196

Alessiass:
Jira task: [Link](https://tracker.intive.com/jira/browse/PATRO23-196)
Auth.js: [Link](https://authjs.dev/)

Requested features:
- Credentials provider should use email+password login
- Storage policy are cookies
- Refresh token integration
- Page securing middleware
- Setup of NEXTAUTH_SECRET

Status:
- Credentials provider added
- No endpoint defined for refresh token integration
- Page securing middleware added - only logged users can browse pages /budgets and /reports
- Login page fully functional with error toast in case of wrong credentials
- Env variables defined and added
- NEXTAUTH_SECRET defined


apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,63 @@
...
+      return session;
+    },
+  },
+  providers: [
+    CredentialsProvider({
+      credentials: {},
+      async authorize(credentials, req) {
+        const { email, password } = credentials as CredentialType;
+        const res = await fetch(
+          "https://inbudget-patronage-api-dev.azurewebsites.net/user/sign-in",
```
Comment (wlechowicz) [inline apps/web/pages/api/auth/[...nextauth].ts:37]: This should use `API_URL` env variable.

apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,63 @@
...
+          {
+            method: "POST",
+            headers: {
+              accept: "text/plain",
+              "Content-Type": "application/json",
+            },
+            body: JSON.stringify({ email, password }),
+          }
+        );
+        const user = await res.json();
```
Comment (wlechowicz) [inline apps/web/pages/api/auth/[...nextauth].ts:47]: Check `res.ok` before parsing because it will throw in the response is not a parseable JSON.

apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,63 @@
...
+      },
+    }),
+  ],
+  session: {
+    strategy: "jwt",
+  },
+  pages: {
+    signIn: "/sign-in/",
+    signOut: "/sign-in",
+    error: '"/sign-in/"',
```
Comment (wlechowicz) [inline apps/web/pages/api/auth/[...nextauth].ts:60]: Make this uniform, second one has no trailing slash, third one has quotes inside quotes.

apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,63 @@
...
+    CredentialsProvider({
+      credentials: {},
+      async authorize(credentials, req) {
+        const { email, password } = credentials as CredentialType;
+        const res = await fetch(
+          "https://inbudget-patronage-api-dev.azurewebsites.net/user/sign-in",
+          {
+            method: "POST",
+            headers: {
+              accept: "text/plain",
```
Comment (wlechowicz) [inline apps/web/pages/api/auth/[...nextauth].ts:41]: Response is definitely not text/plain, but rather application/json. I don't think you need to set this at all.

wlechowicz (changes requested):

apps/web/pages/api/auth/[...nextauth].ts
Comment (Alessiass) [reply apps/web/pages/api/auth/[...nextauth].ts]: Fixed

apps/web/pages/api/auth/[...nextauth].ts
Comment (Alessiass) [reply apps/web/pages/api/auth/[...nextauth].ts]: Fixed

apps/web/pages/api/auth/[...nextauth].ts
Comment (Alessiass) [reply apps/web/pages/api/auth/[...nextauth].ts]: Removed

apps/web/pages/api/auth/[...nextauth].ts
Comment (Alessiass) [reply apps/web/pages/api/auth/[...nextauth].ts]: Done, but the variable is missing in Vercel.

apps/web/pages/api/auth/[...nextauth].ts
Comment (bdeluga) [inline apps/web/pages/api/auth/[...nextauth].ts:1]: I am commenting on whole file so it is in one place. 📖 

- First of all you dont need async callbacks
- `async jwt` does nothing so we can get rid of it
- we could return just data we want not whole essay that backend returns to us
image
(ofc you have to update types acordingly, also retrieved sub, because TS want ID and yells at me, we dont like red in our IDE)
- we can just stick to one decode library 👷 
- storing accessToken in user is quite tricky, tRPC does this better cause it already has Default session for us to extend
image

- Logging In & Logging Off works without these, so i removed them
image

- Since next-auth likes image instead of avatar i just gave up and changed it in types you made 🚀 
image

Results:
image

### Hope it helps 🙏

bdeluga (changes requested):
Couple of tweaks 🚀

bdeluga (approved):
Nice job ! 🚀

apps/web/env.mjs
bdeluga, line 1: API_URL has to stay on client (and has to stay `NEXT_PUBLIC_`), because other components - that will be mainly "use client" sometimes needs to call api and your change will prevent them from it, here's my [conversation ](https://intive-apu8934.slack.com/archives/C04R6LM4JSV/p1682964811546359?thread_ts=1682962537.131399&cid=C04R6LM4JSV)with @wlechowicz for context.

❗ Client env will work on both server and client component

bdeluga (changes requested):

apps/web/env.mjs
Alessiass: I understand but currently the API_URL on Vercel is very specific and will be only used by Next-Auth at this moment. Should I add another one on client and hide the login behind server or should we keep only the client one? Which would be better?

wlechowicz: As confirmed by the client, the backend URL is public and can be exposed. Vercel is already configured for `NEXT_PUBLIC_API_URL`.

Alessiass: Thanks for feedback! Fixed.

bdeluga (approved):
Nice job ! 🥇

turbo.json
```diff
@@ -4,7 +4,8 @@
   "pipeline": {
     "build": {
       "dependsOn": ["^build"],
-      "outputs": ["dist/**", ".next/**", "storybook-static/**"]
+      "outputs": ["dist/**", ".next/**", "storybook-static/**"],
+      "env": ["API_URL"]
```
wlechowicz, line 8:
```suggestion
      "env": ["NEXT_PUBLIC_API_URL"]
```

apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,83 @@
...
+  jwt: {
+    async encode({ token, secret }) {
+      // @ts-ignore
+      return jwt.sign(token, secret);
+    },
+    async decode({ token, secret }) {
+      // @ts-ignore
+      return jwt.verify(token, secret);
+    },
+  },
```
Comment (wlechowicz) [inline apps/web/pages/api/auth/[...nextauth].ts:81]: Is this needed? This looks like the default behavior which I'd assume works out of the box. Question, not a strict change request.

wlechowicz (changes requested):

apps/web/pages/api/auth/[...nextauth].ts
Comment (Alessiass) [reply apps/web/pages/api/auth/[...nextauth].ts:81]: @wlechowicz Currently it's not required. I added it because there was a point in story that we will use NEXT_SECRET env.
As according to: https://next-auth.js.org/configuration/options#override-jwt-encode-and-decode-methods
We can safely remove it.

apps/web/app/(regflow)/sign-in/page.tsx
```diff
@@ -62,12 +64,20 @@ export default function SignInPage() {
 
   return (
     <Form
-      onSubmit={(values) => {
-        values.email === "smutnarzaba@png.pl" && values.password === "frytki123"
-          ? router.push("/home")
-          : setErrMsg(t(form.errorMessage));
+      onSubmit={async (values) => {
+        const result = await signIn("credentials", {
+          email: values.email,
```
AzulloCyll, line 69: You don't need to asign result constant, because you don't using it in future. So just await signin..... will do the job.

apps/web/app/(regflow)/sign-in/page.tsx
```diff
@@ -62,12 +64,20 @@ export default function SignInPage() {
...
+          } else {
+            setErrMsg(t(form.errorMessage));
+          }
+        });
       }}>
-      {({ submit, errors }) => (
+      {({ submit }) => (
         <FormWrapper>
           <form
             onSubmit={(e) => {
```
AzulloCyll, line 83: Line 85

I think you dont need `submit();` here. Submit function is going from Form Component by houseform.

apps/web/app/(regflow)/sign-in/page.tsx
Alessiass: Fixed! 🚀

apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const authOptions: NextAuthOptions = {
+  callbacks: {
+    jwt({ token, account, user }) {
+      if (account) {
+        token.accessToken = (user as DefaultUser).accessToken;
+      }
+      return token;
+    },
+    session({ session, token, user }) {
```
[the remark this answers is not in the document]
Comment (Alessiass) [reply apps/web/pages/api/auth/[...nextauth].ts]: Fixed! 🚀

apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,83 @@
...
+      session.user.accessToken = token.accessToken as string;
+      return session;
+    },
+  },
+  providers: [
+    CredentialsProvider({
+      credentials: {},
+      async authorize(credentials, req) {
```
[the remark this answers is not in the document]
Comment (Alessiass) [reply apps/web/pages/api/auth/[...nextauth].ts]: Fixed! 🚀

apps/web/app/(regflow)/sign-in/page.tsx
```diff
@@ -62,12 +64,20 @@ export default function SignInPage() {
 
   return (
     <Form
-      onSubmit={(values) => {
-        values.email === "smutnarzaba@png.pl" && values.password === "frytki123"
-          ? router.push("/home")
-          : setErrMsg(t(form.errorMessage));
+      onSubmit={async (values) => {
```
zuzannapaw, line 67: Maybe if function passed in props is long, it would be good to move the declaration higher and just pass the function name in `onSumbit` ? What do you think?

zuzannapaw (approved):
Just a small question! 👀 Great job 🚀

PR #59, opened 2023-04-25
fix: currency selector styles in storybook #P23-267 

Vectari:
Hello! This is PR of my task [Fix Currency Selector styles in storybook](https://tracker.intive.com/jira/browse/PATRO23-267).
I think now it looks like in web and looks more like in design (I deleted borders on focus). 

[Check WEB](https://patronage2023-js-dbqlopdja-wlechowicz.vercel.app/)
Budgets > Add new budget > Currency

[Check Storybook](https://patronage2023-js-storybook-jyppd9mtk-wlechowicz.vercel.app/?path=/story/currency-select--select)

Have a nice day! 😃

### DIFF
--- a/packages/ui/CurrencySelect/CurrencySelect.styled.tsx
+++ b/packages/ui/CurrencySelect/CurrencySelect.styled.tsx
@@ -23,21 +23,21 @@ export const SelectTrigger = styled(Select.Trigger)<SelectTriggerProps>`
   padding-right: 10px;
   transition: border-color 200ms ease-out;
 
-  & label {
-    position: absolute;
-    transform: translateY(-24px);
-    font-size: 12px;
-    font-weight: 600;
-    background-color: ${({ theme }) => theme.input.labelBackground};
-    padding-left: 4px;
-    padding-right: 4px;
-  }
-
   :focus {
     transition: border-color 200ms ease-out;
     outline: none;
     border-color: ${({ theme }) => theme.input.focus};
   }
+`;
+
+export const StyledLabel = styled.label`
+    position: absolute;
+    margin-top: -24px;
+    font-size: 12px;
+    font-weight: 600;
+    background-color: ${({ theme }) => theme.input.labelBackground};
+    padding-left: 4px;
+    padding-right: 4px;
 `;
 
 export const SelectIcon = styled(Select.Icon)`
--- a/packages/ui/CurrencySelect/CurrencySelect.styled.tsx
+++ b/packages/ui/CurrencySelect/CurrencySelect.styled.tsx
@@ -59,18 +59,21 @@ export const SelectContent = styled(Select.Content)`
 `;
 
 export const SelectItem = styled(Select.Item)`
+  outline-color: ${({ theme }) => theme.input.focus};
   padding: 16px;
   gap: 8px;
-
+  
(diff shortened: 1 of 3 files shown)


packages/ui/CurrencySelect/CurrencySelect.styled.tsx
```diff
@@ -59,13 +59,13 @@ export const SelectContent = styled(Select.Content)`
...
   padding: 16px;
   gap: 8px;
 
   &:focus {
     color: ${({ theme }) => theme.input.main};
     background-color: ${({ theme }) => theme.currencySelect.focusBackground};
+    outline-color: transparent;
     &:first-child {
       border-radius: 1em 1em 0 0;
     }
```
wyola, line 71: I think this should work a little bit different - hover on mouse and outline/outline+hover on keyboard. Right now there is only hover on keyboard. Our hover color is very light, and not great for people with vision problems, therefore I don't think we should lose outline on keyboard. 

Check out how it works with `IconPicker` ⬇️ it is the same with the `TransactionDropdownMenu` and side navigation. So there should be separate styles for `hover` and for `focus` (you could actually just ~~steal~~ borrow styles from `TransactionDropdownMenu`🤭 since it is also a list and first and last element need slight modifications of border radius in top/bottom corners)

https://user-images.githubusercontent.com/80450315/234507188-74fa68f0-c882-4394-914a-aa0a6c5ee35a.mp4

Vectari: I agree, borders are more visible but in icon picker now you can focus/hover on two icons in the same time. In my opinion it's not looks good. I can back to my old version - no borders on hover and visible borders after use keyboard but in this situation storybook show borders all the time.

wyola: check styles for `TransactionDropdownMenu`, threre is no such problem

Vectari: You are probably right. I add borders later, after work. Thank you for review 😉

Vectari: I added borders on focus and no borders on hover. There is no situation like we talked about (focus and hover in the same time). Only one thing. In web when you use keyboard you have borders and on mouse you dont have but in Storybook there is always border on selected element when you open list. I dont think this is a big problem because this is only in Storybook which i dont know why but realy like borders :D. SB added me borders in the most of the time while I style even if i dont want it... 

I hope is better now 😄

wyola (approved):
LGTM! but maybe don't merge into main until after today's demo 😅 (maybe it doesn't matter and it might be just my trauma after yesterdays and today's build problems 👀)

packages/ui/CurrencySelect/CurrencySelect.styled.tsx
wyola: all nice! ⭐

Vectari:
I wouldn't dare touch anything two hours before the demo 😄 Thanks for approve!

AzulloCyll (approved):

wlechowicz (approved):

bdeluga (approved):
🚀

PR #60, opened 2023-04-26
fix: switch to transient props in Logo component #P23-269

AzulloCyll:
Hello. I fixed problems with our logo wich was generating some console errors. Actually our console is clean. :)

Task: (https://tracker.intive.com/jira/browse/PATRO23-269

Problem is solved by changing original props to transient props in styled components. 
You can read about it [here](https://styled-components.com/docs/api#transient-props)


packages/ui/Logo/index.tsx
```diff
@@ -4,62 +4,56 @@ import Link from "next/link";
...
+const LogoStyledContainer = styled(Link)<LogoProps>`
   font-family: "Signika", sans-serif;
   font-weight: 600;
   text-decoration: none;
   display: flex;
   justify-content: space-around;
   align-items: center;
-  color: ${({ white }) => (white ? "white" : "#1E4C40")};
+  color: ${({ $white }) => ($white ? "white" : "#1E4C40")};
   /*
```
monika-klokosinska-intive, line 26: @AzulloCyll is this comment still relevant? If not, we could remove it.

AzulloCyll: @monika-klokosinska-intive 

It was added to just show logo on dark background in storybook when we had only storybook to show something on demo.

So I've got ridden of it, and added wrapper for the Logo in storybook. It works good :)

packages/ui/Logo/index.tsx
```diff
@@ -4,62 +4,53 @@ import Link from "next/link";
...
   align-items: center;
-  color: ${({ white }) => (white ? "white" : "#1E4C40")};
-  /*
-	temporarily adding background to white version to make it visible on white background
-	*/
-  background: ${({ white }) =>
-    white ? ({ theme }) => theme.logo.main : "none"};
-  ${({ logoWidth }) =>
-    logoWidth &&
+  color: ${({ $white }) => ($white ? "white" : "#1E4C40")};
```
aniasgit, line 25: Why the colors are not from theme?

aniasgit, line 25: Why the colors are not from theme?

AzulloCyll: I'ts old component. I'll fix it of course.

aniasgit (approved):

packages/ui/Logo/index.tsx
```diff
@@ -4,62 +4,52 @@ import Link from "next/link";
...
+      font-size: ${($logoWidth / 6).toFixed(2) + "px"};
+      letter-spacing: ${(($logoWidth / 6) * -0.02).toFixed(2) + "px"};
     `}
 `;
 
-const LogoSignet = styled.span<LogoSignetProps>`
-  background: url(${({ white }) => white ? logoVersions.whiteURL : logoVersions.colorURL})
-    no-repeat center / contain;
+const LogoSignet = styled.span<LogoProps>`
+  background-image: url(${({ $white }) => $white && logoVersions.white || logoVersions.color});
```
Noshi96, line 38: It's kinda hacky condition✍️ . Should also work within the use ternary operator. I think this will be easier to read :D 

```suggestion
  background-image: url(${({ $white }) => $white ? logoVersions.white : logoVersions.color});
```

Noshi96 (approved):
LGTM 🚀

packages/ui/Logo/index.tsx
monika-klokosinska-intive: Good point here by @Noshi96! @AzulloCyll could you apply? Afterwards we could merge this to main :).

AzulloCyll: Aplied, but unfortunately we exceeded our deploy limit on vercel. "...try again in 5 hours..." OMG, heavy penalty.

bdeluga (approved):
Nice job! 🚀 Those red errors had me scared 😱.

packages/ui/Logo/index.tsx
```diff
@@ -4,62 +4,52 @@ import Link from "next/link";
...
+  ${({ $logoWidth }) =>
+    $logoWidth &&
     css`
-      width: ${(logoWidth * 0.25).toFixed(2) + "px"};
+      width: ${($logoWidth * 0.25).toFixed(2) + "px"};
     `}
 `;
 
-export const Logo = ({ white, logoWidth = 144 }: LogoProps) => {
+export const Logo = ({ $white, $logoWidth = 144 }: LogoProps) => {
```
wlechowicz, line 49: That's not how this should work. The props should not be transient "on the outside", as in Logo component should accept `white` and `logoWidth` props, without the `$`. But it should use transient props inside, so that they don't get attached to DOM, so `<LogoSignet $logoWidth={logoWidth} $white={white} />`.

wlechowicz (changes requested):

packages/ui/Logo/index.tsx
AzulloCyll: Ok, changes was made, and commited.

Alessiass (approved):

wlechowicz (approved):

PR #66, opened 2023-04-27
feat: P23-270 Consolidate budgets details and budgets transactions into one page

aniasgit:


apps/web/app/(navigation)/budgets/BudgetsContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <BudgetContentWrapperStyled>
+      <BudgetBasicInformation budget={budget} />
+      <TransactionsTable
+        budget={budget}
+        setSorting={(column) => console.log(column)}
+      />
+    </BudgetContentWrapperStyled>
+  );
+
+  useEffect(() => {
```
monika-klokosinska-intive, line 52: @aniasgit first, I would put all `useEffect` above the jsx (e.g. line 42). This way the code is clearer - you know where `budget` value was decided before it's used in `mainCardContent`.

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+        setBudgets(result.budgets);
+        console.log(result);
+      });
+  }, []);
+
+  useEffect(() => {
+    console.log(id);
+    console.log(budgets);
+    budgets &&
+      budgets.map((currentBudget) => {
```
monika-klokosinska-intive, line 65: @aniasgit if you want to find one element in array, `find` is always a better option. 

`map` is misused here, as the point of `map` is to return a new table based on the old one. You don't return anything here, instead use `map` to call a side effect. If you want to loop through an array and call a side Effect, `forEach` is the right choice.

But again, since we are looking for one element here, `find` is the correct choice. One benefit of `find` is that it will stop the lookup as soon as it found the matching element, whereas `map` or `forEach` would loop through the whole array regardless.

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+        if (currentBudget.id === id) {
+          setBudget(currentBudget);
+          console.log(currentBudget);
```
monika-klokosinska-intive, line 68: @aniasgit finally you can get rid of `budget` in state and the whole second `useEffect` hook, to streamline the behaviour.

```
  const [budgets, setBudgets] = useState<Budget[]>([]);

  useEffect(() => {
    fetch(`../budgets.json`)
      .then((response) => response.json())
      .then((result) => {
        setBudgets(result.budgets);
        console.log(result);
      });
  }, []);

  const budget = budgets.find((currentBudget) => 
    currentBudget.id === id);
```

If you initialise `budgets` with an empty array, you don't have to wait for the fetch to finish to start looking up. `find` returns `undefined` if nothing is found, so your checks for `budget` still work as expected. 

If you don't like the idea of `find` running on every re-render, it can be wrapped in `useMemo` 

```
  const budget = useMemo(() => budgets.find((currentBudget) =>
    currentBudget.id === id), [id, budgets]);
```

Now we arrive at the same behaviour you had before with simpler code and without cascading `useEffect` which are always hard to manage.

This is a great resource about useEffect: https://react.dev/learn/you-might-not-need-an-effect Of course, I don't expect you to read it today, but it's good to keep in mind for the future :).

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
aniasgit: done

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
aniasgit: changed to find

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
aniasgit: done as you advised

PawelPazderski (approved):
Looks and works really nice, one thing to consider for now I would always start the budget from the top of the container it is in: 
image

apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx
```diff
@@ -0,0 +1,130 @@
...
+          />
+          <InfoTile
+            label={t(basicInformation.labels.currency)}
+            dataToRender={currencyInfo}
+          />
+        </TileWrapper>
+      </BasicInfoWrapper>
+    </>
+  );
+}
```
Comment (zuzannapaw) [inline apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx:130]: In an idea of reduce code amount, maybe a good idea would be to make an array full of objects {label,dataToRender}. It would look something like this. What do you think of that? I tried that on web and it worked correctly, plus it gives us less code ☺️ 
image

zuzannapaw (commented):
Just small suggestions from me + I think we have already task for it but I'm not sure, but wen I expand `TransactionDropdownMenu` it suddenly change a bit the size of the page. Anyway great job and congratulations, it looks very nice!

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
```diff
@@ -0,0 +1,57 @@
...
+  justify-content: flex-start;
+  align-items: center;
+  gap: 32px;
+`;
+
+export const BudgetsContent = ({ id }: BudgetsContentProps) => {
+  const [budgets, setBudgets] = useState<Budget[]>([]);
+
+  useEffect(() => {
+    fetch(`../budgets.json`)
```
bdeluga, line 28: This could be just `/budgets.json` since it's in public folder

apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx
Comment (PawelPazderski) [reply apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx:130]: thought of that also, but in the end you write more code here, less there or the other way around.. in this case I decided not to use map as we have only certain amount of information, but I have nothing against changing it to your suggestion

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const BudgetContentWrapperStyled = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
+  align-items: center;
+  gap: 32px;
+`;
+
+export const BudgetsContent = ({ id }: BudgetsContentProps) => {
```
bdeluga, line 24: id could be retrieved using usePathname()  `const id = usePathname().replace('/budgets/',"")` and then we don't need to pass it

in next [13.3](https://beta.nextjs.org/docs/api-reference/use-params) we could just do `const {id} = useParams()` but we are working with 13.1

bdeluga (approved):
Small ideas, overall I am impressed how well you taped all of those components togheter 🎖️ Good Job 🚀

apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx:130]: @PawelPazderski you are right as well, because we have to do additional array anyway. Just thougt that the code in render would be shorter and more transparent. But that was just an idea that had crossed my mind ☺️

Noshi96 (approved):

apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx:130]: I'm analysing code and I thought as well about one thing -> In `BudgetsContent` we look for current budget according to passed `id` from props, and after that we pass that current budget  to `<TransactionsTable>` in it's prop (budget), but actually inside `<TransactionsTable>` we use only id from this budget to get proper transactions```
useEffect(() => {
    fetch(`/budget/${budget.id}.json`) ...
```
So what do you think of changing `<TransactionsTable>` 'budget' prop for currentBudget.id? We could pass `id` directly to the component instead passing whole budget. Then in props it would look smth like `<TransactionsTable id={currentBudget.id}>` and for fetch function in useEffect ` fetch(`/budget/${id}.json`)` .What do you think about it?

File: apps/web/lib/dictionary.tsx [hunk 7]
```diff
@@ -323,17 +322,29 @@ const dictionary = {
...
+        USD: { en: "United States Dollar", pl: "Dolar Amerykański"},
+        PLN: { en: "Polish Zloty", pl: "Polski Złoty"},
+        GBP: { en: "British Pound", pl: "Funt Brytyjski"},
+        EUR: { en: "Euro", pl: "Euro"},
+      },
+      dateFormats: { en: "en-GB", pl: "pl"}
+     },
     tableDates: {
       yesterday: { en: "Yesterday", pl: "Wczoraj" },
       today: { en: "Today", pl: "Jutro" },
```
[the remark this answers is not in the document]
PawelPazderski: image
;)

apps/web/app/(navigation)/SideNavigationBarNavListData.tsx
```diff
@@ -29,19 +29,29 @@ export const BudgetsSubMenuNavListContents = [
...
+    href: "/budgets/36cc0777-2874-4d0c-a389-91280e45e836",
     id: 3,
   },
+  {
+    ComponentToRender: (
+      <>
+        <IconStyled icon="payments" iconSize={24} />
+        <SpanStyled>Praca</SpanStyled>
+      </>
+    ),
```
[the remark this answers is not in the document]
aniasgit: Beacause mock data said that ;)

apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx:130]: @zuzannapaw you can't pass only `id` since `TransactionsTable` uses also `currency` from budget to render `CurrencyAmount` component. So if this change were to happen, there also needs to be some `currency` prop or it can just stay as is.

apps/web/lib/dictionary.tsx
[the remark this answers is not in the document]
wyola: lol, it was late xD

apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx
Comment (aniasgit) [reply apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx:130]: We use also currency from budget, so only id it's not enough. Yesterday it was discussed on slack, I adapted to the version of the table prepared by Ola. It could be changed to just id and currency, but I don't want to mess with the table today.

apps/web/lib/dictionary.tsx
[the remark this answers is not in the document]
aniasgit: someone tired wrote this :)

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
aniasgit: firs comment done, folder name stays as it is. It's also a name of section in submenu, so let's leave it the same.

apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx:130]: @aniasgit true! As @wyola said as well! It was my oversight ☺️

PR #67, opened 2023-05-01
discovery: pagination behaviour #P23-272

zuzannapaw:
- task in jira https://tracker.intive.com/jira/browse/PATRO23-272 
- react paginate on github https://github.com/AdeleD/react-paginate and npm https://www.npmjs.com/package/react-paginate

To use it correctly we need to import `ReactPaginate from "react-paginate"`

`<ReactPaginate>` accepts props that allow us to define how pagination should look and work like:

- `onPageChange` - we can pass function that will be called when user changes the page,
- `pageCount` -  we can determine the amount of pages that will be available,
- `renderOnZeroPageCount` - we can define what will be rendered when there are 0 pages. If we don’t want to render anything, `null` should be passed,
- `breakLabel`, `nextLabel` , `previousLabel` props **are able to accept styled components** as labels for pagination's elements,
- **behaviour of three dots** -> prop `pageRangeDisplayed` allows to define how many pages should be visible before or after break label (...). F.ex: `pageRangeDisplayed={3}` => 3 pages are displayed before and after break label, depends where user click. When user change pages, the position of break label changes as well in proper moment to preserve the `pageRangeDisplayed`, 
- props such as `pageClassName` , `activeClassName`, `nextLinkClassName` etc. allow to pass class names in the aim of customizing pagination's elements (css styling).

Demo is on the main page of  application 👀

**Update** 💥 

1. Kudos @mpiwonski  for comment about disabled buttons.  In PR #71 @Noshi96 already discovered option for making previous/next button disabled. To achieve it, prop `disabledClassName` must be added with our class name for proper styling.
 
2. **Pagination behaviour with more items** (= more pages) -> Pagination can expand up to 10 visible pages. To preserve this, when there is more than 10 pages, two pairs of `...` appear. By clicking it, we can jump over the amount of pages that was set in `pageRangeDisplayed`.

### DIFF
--- a/apps/web/app/(navigation)/(singleCard)/MainPageTranslated.tsx
+++ b/apps/web/app/(navigation)/(singleCard)/MainPageTranslated.tsx
@@ -1,12 +1,15 @@
 "use client";
 
-import { TypoStyled, LinkStyled } from "app/(navigation)/HomePageComponents";
-import { useTranslate } from "lib/hooks";
 import styled from "styled-components";
+import { Pagination } from "app/(navigation)/Pagination";
+
+//simulation of transaction's data
+const items: number[] = [];
+for (let i = 1; i <= 100; i++) {
+  items.push(i);
+}
 
 export const MainPageTranslated = () => {
-  const { dict, t } = useTranslate("MainPage");
-
   const ContentWrapperStyled = styled.div`
     display: flex;
     flex-direction: column;
@@ -15,8 +18,7 @@ export const MainPageTranslated = () => {
 
   return (
     <ContentWrapperStyled>
-      <TypoStyled>{t(dict.welcomeText)}</TypoStyled>
-      <LinkStyled href="/sign-up">{t(dict.createAccountLink)}</LinkStyled>
+      <Pagination data={items} itemsPerPage={10} />
     </ContentWrapperStyled>
   );
 };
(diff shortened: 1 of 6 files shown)


wlechowicz:
```
web:build: Failed to compile.
web:build: ./app/(navigation)/Pagination.tsx
web:build: 60:13 Error: Missing "key" prop for element in iterator  react/jsx-key
```

mpiwonski (commented):
Hi @zuzannapaw, thanks for your work. It seems like you have achieved main objective that is to create a pagination function, whereas there's a few issues I have observed with regards to task requirements in Jira: https://tracker.intive.com/jira/browse/PATRO23-271.

1. It seems like the appeareance of back button does not change whether it's a first page and we cannot go back to the previous one or there's, e.g. 6th page and back button still looks the same (?) - the same situation with forward button

I know it's just POC, but it's still worth mentioning about some missing pieces :)

zuzannapaw (commented):
@mpiwonski  thank you for your remark! 👀  I am definitely going to do some research. I will update results in PR's comment 💬 ☺️

wlechowicz:
I think the goal of this PR has been achieved as it's been used in #71
As this wouldn't get merged on its own, I think you can close the PR and mark the task as done.

wlechowicz (commented):
Approved, but not to be merged.

zuzannapaw (commented):
@wlechowicz thank you for your approve, therefore PR has been closed ☺️

PR #68, opened 2023-05-01
fix: transactions table #P23-282

wyola:
## Fixes for transactions' table
task: https://tracker.intive.com/jira/browse/PATRO23-282
### What changed ✏️ 
- **column sizes are set as %** - there is a difference in columns' width when you compare it to the design. This is due to the fact, that it has to display well on a variety of screen sizes. Setting it to widths similar to the ones on the design would always produce some problems on different widths, especially around breakpoints. I tried to find the best widths that work well on a bigger screen, around 1025px, and down to 768px. Below that we would need additional design and probably some columns should be hidden then.
- **other differences compared to design:**
   - as we discussed, the "Creator" column (header and contents) is centered (it is to the left in the design)
   - I centered also the "Amount" header (which seems to be to the left in the design), so there is a smaller gap between the header and column's contents. It could maybe be to the right, but on wider screens, it did not look good to have such a big white space between the "Status" and "Amount" headers
- **header bottom border** - I added styles to make it look like in the design (full width of the `Card`)
- **fix for table shift** - I fixed the table shift caused by `TransactionDropdownMenu` hiding the scrollbar
### Issues 😢 
It does not look great on a very wide screen. There probably should be some limits to `Card` width on such a screen, or some other way around it, but again, there is no design for that

image

### Things that might look weird but aren't 😅
- negative margins on `TableWrapper` - done to have a bottom border of the table header be of the full width of the `Card`
image

- creating a grouping row adds an empty cell at the beginning of each row - it had to be sized to 1px instead of having `display: none` since the latter would for some reason shift the table layout to the left and add some space on the right 🙄
image

- each column has a commented-out line that adds a background color to it, so it is easier to see how columns resize - I am leaving it for now, will remove it before merging

### DIFF
--- a/apps/web/app/(navigation)/budgets/[id]/TransactionsTable.styled.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/TransactionsTable.styled.tsx
@@ -2,17 +2,22 @@ import styled from "styled-components";
 import { CurrencyAmount } from "ui/CurrencyAmount";
 
 export const TableWrapperStyled = styled.div`
-  width: 100%;
-  display: flex;
-  flex-direction: column;
+  margin-left: -48px;
+  margin-right: -48px;
 
   // table styles
   .ka-table {
-    width: 100%;
+    tr > :first-child {
+      padding-left: 48px;
+    }
+
+    tr > :last-child {
+      padding-right: 0;
+    }
   }
 
   .ka-empty-cell {
-    width: 1px;
+    width: 48px;
   }
 
   // header styles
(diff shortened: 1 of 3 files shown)


wlechowicz (changes requested):
Right-aligned "Amount" cells look bad.

The whole table is unusable on mobile:
obraz

wyola:
> Right-aligned "Amount" cells look bad.
> 
> The whole table is unusable on mobile: 

@wlechowicz After asking clients, I changed "Amount" to be aligned to the left (https://intive-apu8934.slack.com/archives/C04R6LM4JSV/p1683016578103229)

As for the mobile design - I thought we talked that there is no design for those screens, therefore we do not work on it now. Like I wrote in the PR's description:

> I tried to find the best widths that work well on a bigger screen, around 1025px, and down to 768px. **Below that we would need additional design and probably some columns should be hidden then.**

So should I hide some columns below 768px? Set the table to have min-width of that value so there is a horizontal scroll?

Alessiass (approved):
Looks and works really nice! 🚀

wlechowicz (approved):

PR #69, opened 2023-05-01
feat: env stuff 📦 

bdeluga:
## Introducing **[T3.env](https://env.t3.gg)** 👷‍♂️ 

Zod library that help us solve env problems when even rubber duck cannot. ✨ 

## Env schema :page_with_curl:

First we create .mjs file and create our env schema.

 ⚠️Note that demo shows `client-side` env - it has to be prefixed with NEXT_PUBLIC_. 
```tsx
//env.mjs

import { createEnv } from "@t3-oss/env-nextjs";
import { z } from "zod";

export const env = createEnv({
  client: {
    NEXT_PUBLIC_SOME_KEY: z.string().url(),
  },

  runtimeEnv: {
    NEXT_PUBLIC_API_URL: process.env["NEXT_PUBLIC_SOME_KEY"],
  },
});

```
Since we are using next.js `runtimeEnv` is strict by default  and we have to destructure process ourselves.

## Setting env variable 🔨 

```tsx
//.env.local

NEXT_PUBLIC_SOME_KEY="Q2FrZSBpcyBhIGxpZQ=="

```

## Usage  🎯 

```tsx

//Budgets.tsx

import {env} from "@/env.mjs" //same import on both server and client side

const res = fetch(env.NEXT_PUBLIC_SOME_KEY).then(...)

```

## Disclaimer 🚧 

Since every bone of my body stops me from posting **ANY** api url in github. As a good practise let's keep .env.local in `.gitignore` and please add `API_URL` in your .env.local. Dont forget about `NEXT_PUBLIC_`. 🚀

### DIFF
--- a/apps/web/env.mjs
+++ b/apps/web/env.mjs
@@ -1,0 +1,11 @@
+import { createEnv } from "@t3-oss/env-nextjs";
+import { z } from "zod";
+
+export const env = createEnv({
+  server: {
+    API_URL: z.string().url(),
+  },
+  runtimeEnv: {
+    API_URL: process.env.API_URL,
+  },
+});
--- a/apps/web/env.mjs
+++ b/apps/web/env.mjs
@@ -6,7 +6,8 @@ export const env = createEnv({
     NEXT_PUBLIC_API_URL: z.string().url(),
   },
 
-  // need to use [] because process.env works differently in turbo-land
+  // need to use [] because process.env works in next js
+  //https://stackoverflow.com/a/76105098
   runtimeEnv: {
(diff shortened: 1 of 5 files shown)


wlechowicz (changes requested):
LGTM, but please add NEXTAUTH_SECRET env var since we have that too.

wlechowicz (approved):
Never mind, it's still an open PR that env var.

Vectari (approved):

Alessiass (approved):

monika-klokosinska-intive (approved):

PR #70, opened 2023-05-01
feat: edit budget P23-40

mpiwonski:
Hi all,
That wil be a draft pull request for the time being as still there is a few things that need to be discussed.

1. As I mentioned in Slack, there's an issue with **type compatibility for `Budget`**. There's 2 types defined, which are slightly different to each other and each requires some code modifications for an app to work accordingly.

2. There's a create budget endpoint that can already be accessed, whereas there's a question whether we should also have a separate one for editing budget (PATCH?). As for now - budgets state is updated once a specific budget has been changed, so the modifications are immediately presented in the interface.

**Update: 2/05/23**

Backend has been working on edit budget API and still needs some time to complete a task (slack at requirements conversation).


wlechowicz:
Deployment fails due to the type problem mentioned in the description.
obraz

AzulloCyll (commented):
image

That error is showing when you delete a Date. It means that you can't execute function `.getTime()` on null.
You can get my handler for this situation from `createNewBudget` which prevents getTime from calling in certain situation.

image

or just simply get rid of X button from input. But this case should be discussed.

You can show data object wich will be sent to backend in console after you click "Save" button.

apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -139,6 +139,7 @@ export const CreateNewBudget = ({ onClose }: NewBudget) => {
             <form
               onSubmit={(e) => {
                 e.preventDefault();
+                onClose();
```
AzulloCyll, line 142: When you add onClose here, validation will not goes in. 
Check it on AddNewBudgetModal when all fields in form are empty.

AzulloCyll (changes requested):

apps/web/app/(navigation)/CreateNewBudget.tsx
mpiwonski: Thanks for your insights. Fixed.

mpiwonski:
> image
> 
> That error is showing when you delete a Date. It means that you can't execute function `.getTime()` on null. You can get my handler for this situation from `createNewBudget` which prevents getTime from calling in certain situation.
> 
> image
> 
> or just simply get rid of X button from input. But this case should be discussed.
> 
> You can show data object wich will be sent to backend in console after you click "Save" button.

Thanks for your insights. Eventually I combined your Budget type with @bdeluga Budget type, which also implied a need to change one of the functions related to timestamp in BudgetBasicInfo, and now it seems it works appropriately :)

AzulloCyll (approved):

packages/ui/EditIcon/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const EditIconStyled = styled.span`
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  height: 2.5em;
+  width: 2.5em;
+  border-radius: 8px;
+  background-color: #f7f7f7;
+  color: #7e7e7e;
```
aniasgit, line 16: add colors from theme

packages/ui/index.tsx
```diff
@@ -1,3 +1,5 @@
+import { EditIcon } from "./EditIcon";
```
aniasgit, line 1: You don't need to import it here.

packages/ui/EditIcon/index.tsx
```diff
@@ -0,0 +1,29 @@
+import styled from "styled-components";
+import { Icon, IconType } from "../Icon";
+
+interface EditIconInterface {
+  onClick: () => void;
+}
+
+const EditIconStyled = styled.span`
```
aniasgit, line 8: If we want to click on it, there should probably be a button.

apps/web/app/(navigation)/EditBudget.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+export const EditBudget = ({
+  budget,
+  onClose,
+  handleHideEditBudgetModal,
+  handleBudgetsEdit,
+}: EditBudgetProps) => {
+  const [editBudget, setEditBudget] = useState(budget);
+  const [selectedIcon, setSelectedIcon] = useState(budget.icon);
+  const [defaultValueTabs, setDefaultValueTabs] = useState("settings");
```
Noshi96, line 44: I think we don't need to useState for defaultValue, because it appears only once, and we don't expect its value to change dynamically. You can simply write `const defaultValueTabs = "settings"` above the component declaration imo :D

packages/ui/EditIcon/index.tsx
mpiwonski: Thanks for your insight. Fixed :)

apps/web/app/(navigation)/EditBudget.tsx
mpiwonski: Fair point :)

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
```diff
@@ -38,13 +41,36 @@ export const BudgetsContent = () => {
     [id, budgets]
   );
 
+  const showEditBudgetModal = () => {
```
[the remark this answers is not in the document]
mpiwonski: Right, but showEditBudgetModal/hideEditBudgetModal are used in different places with, and I think it's good to leave them separated, and each component receives a specific method according to their needs and it's semanticly clear.

packages/ui/EditIcon/index.tsx
```diff
@@ -0,0 +1,29 @@
+import styled from "styled-components";
+import { Icon, IconType } from "../Icon";
```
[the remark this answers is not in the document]
mpiwonski: Done.

packages/ui/index.tsx
mpiwonski: It was by mistake. Fixed.

packages/ui/EditIcon/index.tsx
mpiwonski: Done.

apps/web/lib/dictionary.tsx
```diff
@@ -284,11 +279,97 @@ const dictionary = {
...
+      },
+      endDate: {
+        en: "End date",
+        pl: "Data końcowa",
+      },
+    },
+    errors: {
+      min3characters: {
+        en: "Budget name must have at least 3 characters.",
+        pl: "Nazwa musi mieć conajmniej 3 znaki.",
```
wlechowicz, line 294:
```suggestion
        pl: "Nazwa musi zawierać co najmniej 3 znaki.",
```

apps/web/lib/dictionary.tsx
```diff
@@ -284,11 +279,97 @@ const dictionary = {
...
+  },
+  EditBudgetModal: {
+    title: {
+      en: "Edit budget",
+      pl: "Edytuj budżet",
+    },
+    tabs: {
+      settings: {
+        en: "Settings",
+        pl: "Ustawiena",
```
wlechowicz, line 334:
```suggestion
        pl: "Ustawienia",
```

wlechowicz (changes requested):
Overall, in EditBudget component, there's too much inline logic in JSX and most of it should be placed above, so that the JSX (return) focuses more on the layout, and the business logic - functions, evaluations - are in the body of the component, before the return.

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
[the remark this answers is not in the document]
Noshi96: Yep I agree with you :D

mpiwonski (commented):
Thanks @wlechowicz for all your remarks and suggestions. I have included them in recent code changes. I have also got rid of `editBudget` state as now we only depend on the values avaiable after form submission. It definitely made code more neat and functional. Please take a look at latest changes and let me know if it's OK or there's anything I could improve :)

packages/ui/EditIcon/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export const EditIcon = ({ onClick }: EditIconInterface) => {
+  return (
+    <EditIconStyled onClick={onClick}>
+      <Wrapper>
+        <Icon icon="edit" />
+      </Wrapper>
+    </EditIconStyled>
+  );
+};
```
wyola, line 36: you could cut this code a bit ✂️ 
- with few changes to `EditIconStyled` you don't need the `Wrapper`
- `Icon` has a `color` prop that you could use

```suggestion
const EditIconStyled = styled.button<EditIconInterface>`
  cursor: pointer;
  border: none;
  background-color: ${({ theme }) => theme.editIcon.background};
  height: 2.5em;
  width: 2.5em;
  border-radius: 8px;
  flex-shrink: 0;
  line-height: 0;
`;

export const EditIcon = ({ onClick }: EditIconInterface) => {
  const theme = useContext(ThemeContext);
  return (
    <EditIconStyled onClick={onClick}>
        <Icon icon="edit" color={theme.editIcon.main}/>
    </EditIconStyled>
  );
};
```
`flex-shrink` is there so that button does not shrink when page is resized ⬇️ (now I actually noticed that `BudgetIcon` has this issue 👀 )
 
image

PS you have to add to imports to use color prop
`import styled, { ThemeContext } from "styled-components";`
`import { useContext } from "react";`
and to component
`const theme = useContext(ThemeContext);`

apps/web/app/(navigation)/EditBudget.tsx
```diff
@@ -0,0 +1,243 @@
...
+    } else if (dateType === "end-date")
+      return budget.endDate ? new Date(budget.endDate) : null;
+  };
+
+  const getSelectedValue = (value: Date | null) =>
+    value ? new Date(value) : null;
+
+  const checkEndDateOnChangeValidate = (
+    val: Date | null,
+    form: FormInstance<any>
```
zuzannapaw, line 60: Could you tell me why did you use here `any` type? 🤔

apps/web/app/(navigation)/EditBudget.tsx
```diff
@@ -0,0 +1,243 @@
...
+                          />
+                          <DatePickerErrorStyled>
+                            {errors[0]}
+                          </DatePickerErrorStyled>
+                        </DatePickerWrapperStyled>
+                      )}
+                    </Field>
+                  </InputWrapperFullFlex>
+                </Tabs.Content>
+                <Tabs.Content value="share">Welcome to share</Tabs.Content>
```
zuzannapaw, line 229: I think this should come from dictionary as well 😃

zuzannapaw (commented):
Good job! Although I'm wondering why the modal does not show up in the middle of the page, but it's moved a bit on the left. 
image

apps/web/app/(navigation)/EditBudget.tsx
AzulloCyll: No, please don't change this. It's only placeholder!! :)

zuzannapaw: Okay!! @AzulloCyll

apps/web/app/(navigation)/EditBudget.tsx
```diff
@@ -0,0 +1,243 @@
...
+    const start = val! && val.getTime();
+    const end =
+      form.getFieldValue("start-date")!.value &&
+      form.getFieldValue("start-date")!.value.getTime();
+
+    if (start && end) {
+      if (start < end) return Promise.reject(t(dict.errors.dateBeforeStart));
+    }
+    return Promise.resolve(true);
+  };
```
wyola, line 71: I think that there are a few things here that are a little confusing, because of the variables' names used 👀 

For example, there is `const end` but it regards `start-date`, and `start` variable is really the timestamp of end date. As a result, because of the current names of the variables, the "if" statement: `if (start < end)` at first glance makes you wonder why there is an error for a situation when `start < end`.

I also think that you can simplify code for `start` and `end` variables using `?` operator.

To sum up, after changing variables' names and mentioned above simplifaction you could arrive at this:

```suggestion
  const checkEndDateOnChangeValidate = (
    endDateValue: Date | null,
    form: FormInstance<any>
  ) => {
    const endTimestamp = endDateValue?.getTime();
    const startTimestamp = form.getFieldValue("start-date")!.value?.getTime();

    if (endTimestamp && startTimestamp) {
      if (startTimestamp > endTimestamp) return Promise.reject(t(dict.errors.dateBeforeStart));
    }
    return Promise.resolve(true);
  };
```

apps/web/app/(navigation)/EditBudget.tsx
mpiwonski: Hi @zuzannapaw, I've been trying to look up default type of `(val, form)` in houseform docs, but it seems I cannot find them, however it was suggested in VSCode, so the types would be compatible and there would not be TS error.

apps/web/app/(navigation)/EditBudget.tsx
```diff
@@ -0,0 +1,243 @@
...
+                        <IconPicker
+                          defaultIcon={value}
+                          icons={icons}
+                          onSelect={(icon) => setValue(icon)}
+                        />
+                      )}
+                    </Field>
+                  </IconPickerStyled>
+                  <InputWrapperFullStyled>
+                    <Field
```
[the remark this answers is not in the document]
mpiwonski: Hi @wyola, honestly, I do not have anything against the current way it is displayed now, especially it's also present in create budget form that was created by @AzulloCyll, so he should also be included in the conversation.

PawelPazderski (approved):
LGTM to merge and continue from here with BE integration 🚀

AzulloCyll:
@PawelPazderski 
Description is not required so can be empty. :)

Alessiass (changes requested):

apps/web/app/(navigation)/EditBudget.tsx
```diff
@@ -0,0 +1,227 @@
...
+              endDate: values["end-date"].getTime(),
+            };
+
+            console.log(newBudgetFromValues);
+          }}>
+          {({ submit }) => (
+            <form
+              onSubmit={(e) => {
+                e.preventDefault();
+              }}>
```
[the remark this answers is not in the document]
AzulloCyll: done :)

Alessiass (approved):

apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -362,17 +362,17 @@ export const CreateNewBudget = ({ onClose }: NewBudget) => {
...
                         newBudget.dateEnd ? new Date(newBudget.dateEnd) : null
                       }
                       onSubmitValidate={checkDate}
                       onChangeValidate={(val, form) => {
                         const start = val! && val.getTime();
                         const end =
-                          form.getFieldValue("date-start")!.value &&
-                          form.getFieldValue("date-start")!.value.getTime();
+                          form.getFieldValue("start-date")!.value &&
+                          form.getFieldValue("start-date")!.value.getTime();
```
aniasgit, line 375: Since we use ISOString to communicate with BE, getTime() is no longer needed. DatePicker returns values, in the Date type, which you can compare, and before sending to the backend, you just need to convert them to ISOString (toISOString()).

aniasgit (commented):
I believe all date operations should be changed, we don't need dates in timestamps anymore. All operations on dates can be performed on Date objects returned to us by DatePicker, and before sending to BE, the values should be converted to ISOString. Similarly, date values that we get from BE should be converted to Date for further operations on them.

aniasgit (approved):

apps/web/app/(navigation)/EditBudget.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+    if (startDateValue && endDateValue) {
+      if (startDateValue > endDateValue)
+        return Promise.reject(t(dict.errors.dateBeforeStart));
+    }
+    return Promise.resolve(true);
+  };
+
+  return (
+    <Modal header={t(dict.title)} onClose={() => onClose()} fullHeight>
```
wlechowicz, line 62:
```suggestion
    <Modal header={t(dict.title)} onClose={() => onClose && onClose()} fullHeight>
```

wlechowicz (approved):
LGTM overall, one refactor to be made based on inline comment.

PR #71, opened 2023-05-03
feat: pagination component #P23-271

Noshi96:
## Pagination Component

Ticket: https://tracker.intive.com/jira/browse/PATRO23-271

## Acknowledgements 🎉 

Thanks to @zuzannapaw for preparing the PoC and insight into react-pagination in [#PR](https://github.com/intive/patronage2023-js/pull/67) 🎉 

## Check out this demo 👀  

https://patronage2023-js-web-git-feature-p23-271-pagi-9b2431-wlechowicz.vercel.app/demo

### Usage 
```tsx
<Pagination
  pageSizeOptions={pageSizeOptions}
  onChangePaginationState={onChangePaginationState}
  numberOfPages={Math.ceil(data.length / rowsPerPage)}
/>;
```
- numberOfPages to set the current number of pages

- pageSizeOptions example:
```tsx
const pageSizeOptions = [10, 15, 20, 30, 50];
```

- onChangePaginationState example:
```tsx
const onChangePaginationState = ({
  rowsPerPage,
  currentPage,
}: PaginationStateType) => {
  console.log({ rowsPerPage, currentPage });
};
```
🎯  With this function `onChangePaginationState`, we can lift information about `rowsPerPage` and `currentPage` up, and send a request to the backend
- PaginationStateType:
```tsx
type PaginationStateType = {
  rowsPerPage: number;
  currentPage: number;
};
```
### Usage 
```tsx
<Pagination
  pageSizeOptions={pageSizeOptions}
  pageIndex={pageIndex}
  numberOfPages={Math.ceil(data.length / pageSize)}
  currentPageSize={pageSize}
  onChangePageIndex={onChangePageIndex}
  onChangePageSize={onChangePageSize}
/>;
```
🎯   It gets additional props, which allows us to set the current page, the current page size and functions to handle setting the new pagination state

## Rows Per Page Select 
Inside the pagination, there is a new component for selecting rows on a page from a drop-down
```tsx
<RowsPerPageSelect
  value={rowsPerPage}
  id="rows-per-page"
  label="select-rows-per-page"
  onValueChange={comboBoxChangeOnChangeHandler}
  pageSizeOptions={pageSizeOptions}
/>;
```
- value: Initial select input  `(string)`
- onValueChange example: 
```tsx
const comboBoxChangeOnChangeHandler = (newRowsNumber: string) => {
  //...
};
```

## Worth mentioning
 - ⚠️  Pagination is indexed from 0. By setting currentPage = 0 it will be 1 for the user on the page
 - ~~⚠️ When the number of rows on a page are changed, the current page is reset to 0 (This does not apply to react-paginate, because we cannot set the current page there)~~
- 🛑  Since the internal state of the component has been removed, it is the responsibility of the TableController to reset the page when the page size change
 - ⚠️ We may want to get the total number of pages from backend after sending the request with currentPage and rowsPerPage otherwise we need to make this equation:
```tsx
Math.ceil(data.length / pageSize);
```


zuzannapaw (commented):
I really enjoyed that demo, I'm sure it was a really hard work! From all pagination options I would choose the 3rd one. I think the behaviour of three dots is the best in this version (with two pages visible at the end). My small suggestion would by about visibility of 3 dots inside clicked input. What do you think about hiding it, when input is on focus? 👀  Well done! 🚀

Noshi96:
> 

@zuzannapaw Thanks for your comment! I think we could apply your suggestion here and hide the 3 dots on focus 🚀

Alessiass (approved):
Looks and works nice! I would chose V3.

wlechowicz (commented):
I like v1 the most. We don't need that input, there wouldn't be tens of pages of transactions per budget. If there's really a lot (hundreds) of them, user would switch to 25 rows per page and would need like 5 pages for 100+ transactions. At that point search, sort and filter play a bigger role than jumping to arbitrary pages.
What the view really lacks is filtering by date range, but that's not a feature that has been requested.

For final approval, a choice needs to be made and the demo needs to be removed for he PR to be merged.

bdeluga (commented):
I would choose v1, simply because it's working the way we want. I don't know a single site where by clicking `...` it creates input of some sort. Maybe we could leave three dots and use it as a divider with no usage whatsoever.

PawelPazderski (approved):
Great demo 🥇 , awesome work! I would go for V1

Noshi96:
> 



> I would choose v1, simply because it's working the way we want. I don't know a single site where by clicking `...` it creates input of some sort. Maybe we could leave three dots and use it as a divider with no usage whatsoever.

@bdeluga Here `...` input for you 😄 

image

wyola (commented):
I like the idea with input, but v1 is also good if this is what the majority think. One thing I would consider changing is the behavior when we reach the last page. There are three dots between pages 9 and 10 and it looks a bit strange and three dots don't do anything there. Maybe they could disappear or move to the beginning (or to the left)? 👀 

image

AzulloCyll:
v1

Noshi96:
> I like the idea with input, but v1 is also good if this is what the majority think. One thing I would consider changing is the behavior when we reach the last page. There are three dots between pages 9 and 10 and it looks a bit strange and three dots don't do anything there. Maybe they could disappear or move to the beginning (or to the left)?

@wyola Thank you for your comment! 
I would prefer to hide it when we get to the last pages, because moving it to the beginning raises a question about behavior. If it's on the left, should it take us 3 pages back? That is, when to display it on the left and when to display it on the right. This requires maintaining an additional state about the previous step. If the user moved backward and forward in the last step. There is much more logic to add, I don't know if we need this functionality of moving the dots to the beginning.

aniasgit (approved):
In general, I am also for v1, although I have similar doubts to @wyola , those three dots between the penultimate and the last page look strange.

mpiwonski:
Wow. That's remarkable how much effort you have put into preparing such a demo. I really appreciate that you've presented pros and cons straightaway, so it's easier for us to make some initial decision on which one to pick. Looking from function perspective of component, I would definitely go with **v1** as it works fine with possibilites to jump 3 pages forward with ... and 2 pages backwards with page button, whereas giving a specific page to go to seems unnecessary to me. One more time, good job! :)

Vectari:
V1 looks and works best in my opinion. Great demo!

apps/web/lib/dictionary.tsx
```diff
@@ -471,6 +471,28 @@ const dictionary = {
...
+  Pagination: {
+    pageAriaLabel: { en: "Page", pl: "Strona" },
+    currentPageAriaLabel: {
+      en: "is your current page",
+      pl: "to twoja bierząca strona",
+    },
+    nextPageAriaLabel: { en: "Next page", pl: "Następna strona" },
+    previousPageAriaLabel: { en: "Previous page", pl: "Poprzednia strona" },
+    rowsPerPageText: { en: "Rows per page", pl: "Wierszy na stronie" },
+    breakElementAriaLabel: {
```
monika-klokosinska-intive, line 483: @zuzannapaw Any change you could provide french values for these objects?

apps/web/components/Pagination/RowsPerPageSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (isOpen) {
+      const body = document.body;
+      const bodyWidth = document.body.clientWidth;
+      if (bodyWidth > 768) {
+        body.style.right = "8.5px";
+      }
+    }
+  }, [isOpen]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
Noshi96: @monika-klokosinska-intive thank you for this comment! 

I was able to solve it, but first I want to say that there was a problem with radix select, because when we open the select options radix uses a portal to display the items and also removes the scrollbar, which causes content width jumps. 

There was a similar problem with opening 3 dots in a table. 3 dots is a dropdown and has support for modal mode which can prevent scroll removal. In select we have no such support https://github.com/radix-ui/primitives/issues/1496. 

Currently, in the body in the CSS condition I used a solution that forces the scrollbar to be displayed on wide resolutions and when the height of the content is greater than 100% viewport. I know that !important is not a good practice, but without it, radix forces the scrollbar to be removed. 

```css
body:has(div[data-radix-popper-content-wrapper]) {
  @media only screen and (min-width: 768px) {
    @media only screen and (min-height: 100vh) {
      overflow-y: scroll !important;
    } 
  } 
}
```

It also solves the currency selection problem.

apps/web/app/css/global.css
```diff
@@ -10,6 +10,14 @@ html {
   height: 100%;
 }
 
+body:has(div[data-radix-popper-content-wrapper]) {
+  @media only screen and (min-width: 768px) {
+    @media only screen and (min-height: 100vh) {
+      overflow-y: scroll !important;
+    } 
+  }
```
wyola, line 18: I think that media query can be combined into a single expression:

```suggestion
  @media only screen and (min-width: 768px) and (min-height: 100vh) {
    overflow-y: scroll !important;
  } 
```

apps/web/components/Pagination/RowsPerPageSelect.tsx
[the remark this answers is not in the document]
Noshi96: Okey I see the problem here, actually we can't check if there is a scrollbar on the page, so nested media doesn't work ;/

apps/web/lib/hooks/useHasScrollBar.tsx
```diff
@@ -0,0 +1,25 @@
...
+let timeoutId: NodeJS.Timeout;
+
+export const useHasScrollBar = () => {
+  const [hasScrollbar, setHasScrollbar] = useState<boolean>(false);
+
+  useEffect(() => {
+    if (isMounted) {
+      // prevent execution of previous setTimeout
+      clearTimeout(timeoutId);
+      timeoutId = setTimeout(() => {
```
monika-klokosinska-intive, line 13: @Noshi96 could you explain what's the need for the timeout is? This is some heavy guns you are pulling out here to make the styling work ;).

monika-klokosinska-intive (approved):

PR #74, opened 2023-05-05
feat: category map #P23-287

wyola:
## Category map 🗺️ 
task: https://tracker.intive.com/jira/browse/PATRO23-287
- `categoryMap` maps category names accepted by the backend to an object of type:
```
export type CategoryType = {
  name: string;
  icon: {
    name: string;
    foreground: string;
    background: string;
  };
};
```
- the same type is used by `CategoryIcon` component
- missing colors were added to the theme

### DIFF
--- a/apps/web/lib/category-map.ts
+++ b/apps/web/lib/category-map.ts
@@ -1,0 +1,39 @@
+import { CategoryMap } from "./types";
+import { theme } from "ui/theme";
+
+const categoryMap: CategoryMap = {
+  HomeSpendings: {
+    name: "Home spendings",
+    icon: {
+      name: "home",
+      foreground: theme.categoryIcons.homeSpendings.foreground,
+      background: theme.categoryIcons.homeSpendings.background,
+    },
+  },
+  Subscriptions: {
+    name: "Subscriptions",
+    icon: {
+      name: "subscriptions",
+      foreground: theme.categoryIcons.subscriptions.foreground,
+      background: theme.categoryIcons.subscriptions.background,
+    },
+  },
+  Car: {
+    name: "Car",
+    icon: {
+      name: "directions_car",
+      foreground: theme.categoryIcons.car.foreground,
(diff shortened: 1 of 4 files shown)


aniasgit (approved):

packages/ui/CategoryIcon/index.tsx
```diff
@@ -3,7 +3,6 @@ import { Icon } from "ui";
 import { IconType } from "ui/Icon/index";
 
 type categoryType = {
```
wlechowicz, line 5: This should be the same type, that's shared and re-used across.

wlechowicz (changes requested):

apps/web/lib/category-map.ts
```diff
@@ -0,0 +1,39 @@
+import { CategoryMap } from "./types";
+import { theme } from "ui/theme";
+
+const categoryMap: CategoryMap = {
+  HomeSpendings: {
+    name: "Home spendings",
+    icon: {
+      name: "home",
+      foreground: theme.categoryIcons.homeSpendings.foreground,
+      background: theme.categoryIcons.homeSpendings.background,
```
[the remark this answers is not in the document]
wyola: ok, fixed

packages/ui/CategoryIcon/index.tsx
```diff
@@ -1,19 +1,10 @@
...
-type categoryType = {
-  id: number;
-  name: string;
-  icon: {
-    name: string;
-    foreground: string;
-    background: string;
-  };
-};
+import { CategoryType } from "../../../apps/web/lib/types"
```
wlechowicz, line 4: Nope, that's a sin. Package must not import from an app.
So you have to move the shared type into the package, so that the app can import both the component and the type.

wlechowicz (changes requested):

wlechowicz (approved):

PR #75, opened 2023-05-05
fix: zod validation schema moved out of form for clarity #P23-148

Alessiass:
Improvement based on suggestion from HouseForm creator


bdeluga (commented):
Who am I to disapprove fixes suggested by the author of HouseForm? 🚀

For real now, does this changes how component works in any way or does this change just makes it look prettier.

bdeluga (approved):

PawelPazderski (approved):
LGTM, less code in JSX 👍

PR #77, opened 2023-05-07
fix: table mobile view  #P23-293 

wyola:
## Mobile view for transactions' table
task: https://tracker.intive.com/jira/browse/PATRO23-293
I tried my best to make it usable, but side menu takes a lot of space 😢 
### What changed ✏️ 
- I added a breakpoint, but it does not come from our `media-queries`. It is a breakpoint that worked best for the table (in particular for moving the sorting icon down). I just hardcoded it in table styles, I don't think there is a point in adding it to global `media-queries` (if I am mistaken, feel free to correct me 😅)
- `Status` column was removed (both for mobile and desktop) - backend does not provide it
- `Creator` column is hidden on mobile
- `Card` horizontal paddings are smaller for mobile view
- `ellipsis` for "Name" column
- "Amount" column may seem to be too wide in some cases, but it has to fit long numbers
- columns' headers and week days come from dictionary now (as they should have from the beginning 🫥)
### Issue 🔴 
I did not find a way to fit such a long number for 375px wide screen ⬇️ last column has a fixed width of a `TransactionDropdownMenu` component and making it smaller messes the layout a bit. Removing `padding-right` on the "Amount" cells also does not help🙄 I don't know if this is very important, since such numbers are rather rare (at least on my bank account 😭 ), but if you can think of anything then let me know!

new-version

### Additional change
- I removed the `media` file and corrected two imports and code that were using it


AzulloCyll (approved):

zuzannapaw (commented):
About this problem with amount column's content when value is really high: maybe we could change font-size from 12px to 10px. It's a bit small for mobile view, but it does not shrink on 375px. 
image

Above 600px rows with long description look like they are not having enough space (the descirpion is very close to the amount). Also maybe adding some padding above this width would make it better? 
image

Good job! 🚀

apps/web/lib/dictionary.tsx
```diff
@@ -345,9 +345,24 @@ const dictionary = {
       },
       dateFormats: { en: "en-GB", pl: "pl" },
     },
-    tableDates: {
-      yesterday: { en: "Yesterday", pl: "Wczoraj" },
-      today: { en: "Today", pl: "Dzisiaj" },
+    transactionsTable: {
+      groupRowDays: {
+        yesterday: { en: "Yesterday", pl: "Wczoraj" },
+        today: { en: "Today", pl: "Dzisiaj" },
```
monika-klokosinska-intive, line 351: @wyola any change we could add the `fr` values for `today` and `yesterday`? @zuzannapaw could you help?

Looks great otherwise!

wyola: @monika-klokosinska-intive you are right, I added other things in french but forgot about these two 🫠 will add it later!

AzulloCyll:
Please resolve conflicts by maerging main to your branch :)

apps/web/lib/dictionary.tsx
zuzannapaw: @wyola  When I was adding french I did not add any fields in dictionary for transactions. But I can help of course! ☺️

Alessiass (approved):
Looks good :D

PawelPazderski (approved):
Looks nice and sure it was not an easy job to fit that nicely. If there is any chance that looks so tight: 
image

other than that, great for me.

bdeluga (approved):
This task made you squeeze water out of stone, and you made that happen... Good job ! 🥇

apps/web/lib/dictionary.tsx
wyola: @monika-klokosinska-intive fixed! @zuzannapaw if you could kindly take a look if it is correct 👀 

```
yesterday: { en: "Yesterday", pl: "Wczoraj", fr: "Hier" },
today: { en: "Today", pl: "Dzisiaj", fr: "Aujourd'hui" },
```

monika-klokosinska-intive:
@wyola please resolve conflicts and we are good to go!

wyola:
> Looks nice and sure it was not an easy job to fit that nicely. If there is any chance that looks so tight Category/Name: image
> 
> other than that, great for me.

I added there a tiny bit, but that's it, since "Category" is already too wide on bigger screens 👀

apps/web/lib/dictionary.tsx
wyola: nvm, it was on main 😄

PR #78, opened 2023-05-07
feat: button with dropdown #P23-290

PawelPazderski:
## Dropdown button component for creating new income & expanse (budget page)

Ticket: 
https://tracker.intive.com/jira/browse/PATRO23-290

The component accepts a label and items as props.
Items need to be an array of objects containing a label and function to execute for each dropdown item.

The button used in the component is not the button from UI, the component didn't cooperate with the DropdownMenu.Trigger from Radix which uses a native button. Tried to find eventual solution here, use ref and forward it down the Button component which i thought would do but I was not successful unfortunately, lost lots of time so I decided to use a native button as in example and give it our styling. Maybe someone will now how to change it eventually, if so feel free to request changes.

### Several changes to styling concerning mobile view also provided here (spaces between elements, paddings, font characteristics etc.):

- BudgetBasicInformation & InfoTile
- Budget Contents
- Card ( HomePageComponents)

The dictionary for the component is also ready to be used by the parent component which will display the button.


bdeluga (approved):
Nice idea to strech InfoTiles to full width, it looks much better that way, also dropdown looks pretty! 🚀 

Can't quite see focused styles, because of black border on dark green color. On the other hand i belive there is no design for that 😞

AzulloCyll:
Button in mobile view should have more padding on left. Maybe 12px, it looks better then. :) But this is a small change only...
Overall work is good and looks good.

apps/web/lib/dictionary.tsx
```diff
@@ -345,6 +345,11 @@ const dictionary = {
       },
       dateFormats: { en: "en-GB", pl: "pl" },
     },
+    createButton: {
+      label: { en: "Create", pl: "Utwórz"},
+      newIncome: { en: "New income", pl: "Nowy dochód"},
```
aniasgit, line 350: pl should be "Nowy wpływ" or "Nowy przychód" I think the secon sounds better but it isn't "dochód", sorry I'm an accountant ;)

packages/ui/ButtonWithDropdown/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+    border: 1px solid transparent;
+  }
+`;
+
+export const ButtonWithDropdown = ({ label, items }: ButtonWithDropdownProps) => {
+  const [isOpen, setIsOpen] = useState(false);
+
+  return (
+    <>
+      <DropdownMenu.Root modal={false} open={isOpen} onOpenChange={setIsOpen}>
```
aniasgit, line 97: If everything is in DropdownMenu.Root, the Fragment is unnecessary.

aniasgit (approved):
I asked for some minor changes, but overall good job.

packages/ui/ButtonWithDropdown/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+    border-top-left-radius: 16px;
+    border-top-right-radius: 16px;
+  }
+  &:last-child {
+    border-bottom-left-radius: 16px;
+    border-bottom-right-radius: 16px;
+  }
+
+  &:focus {
+    border-color: ${({ theme }) => theme.transactionDropdownMenu.outlineFocus};
```
aniasgit, line 78: I'd rather set the color like in CurrencySelect

packages/ui/ButtonWithDropdown/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+    outline: 0;
+  }
+
+  &:hover {
+    background-color: ${({ theme }) =>
+      theme.transactionDropdownMenu.activeBackground};
+    outline: 0;
+    border: 1px solid transparent;
+  }
+`;
```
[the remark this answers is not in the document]
PawelPazderski: Right, the outline here is redundant. The focus outline is default here as in the Button from UI but it is a good idea to make it stand out more definitely.

packages/ui/ButtonWithDropdown/index.tsx
PawelPazderski: Right indeed :)

packages/ui/ButtonWithDropdown/index.tsx
PawelPazderski: Definitely right, I was referring to styles from TransactionDropdown here, but the focus from the CurrencySelect fits as if he had always been here, thanks

apps/web/app/(navigation)/HomePageComponents.tsx
```diff
@@ -1,5 +1,6 @@
 "use client";
-import { device } from "lib/media";
+// import { device } from "lib/media";
```
monika-klokosinska-intive, line 2: @PawelPazderski if you don't need this import anymore, it's better to remove it than comment it!

monika-klokosinska-intive (approved):

apps/web/app/(navigation)/HomePageComponents.tsx
PawelPazderski: of course, I overlooked it, thanks

wlechowicz:
image

PawelPazderski:
> image

ok już patrzę coś mi chyba nie pykło z tym rozwiązaniem konfliktu ;(

PR #79, opened 2023-05-07
feat: budgets page #P23-288

bdeluga:
### Overview 📖 
[Task](https://tracker.intive.com/jira/browse/PATRO23-288) was to implement fetching budget data based on id from params and pass it down to [Flow Controller](https://tracker.intive.com/jira/browse/PATRO23-289). I also added skeleton loading to BudgetBasicInformation while data is loading.

## What's new 👀 
- [Next 13.4.1](https://nextjs.org/blog/next-13-4) - App is now stable and recommanded way of using next.js. Nice 👍 
- [react-loading-skeleton](https://www.npmjs.com/package/react-loading-skeleton) 💀 lightweight library that helps create suspense animations.
- `BudgetBasicInformation` now renders suspensed version while there is no budget data

## Work In Progress 🚧 
Some of features are still missing, [next-auth](https://github.com/intive/patronage2023-js/pull/58) and actual displayment of budgets from backend, so I am using static data to achieve results of task.

## Styling 💅
Here's an extra for those who have any feedback about skeleton loading [thread](https://intive-apu8934.slack.com/archives/C04R6LM4JSV/p1683367957397059).


monika-klokosinska-intive:
@bdeluga could you try to figure out why the web app is not building? We cannot merge otherwise.

wlechowicz:
image

bdeluga:
Right, about that error, will fix it ASAP.

bdeluga:
@wlechowicz @monika-klokosinska-intive error free 🚀

packages/ui/SkeletonLoading/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  width,
+  height,
+  circle,
+}: SkeletonLoadingProps) => {
+  return (
+    <Skeleton
+      width={width}
+      height={height}
+      circle={circle}
+      baseColor="#adaeb3"
```
wyola, line 19: Nice touch adding the skeleton! 🚀 

I would change the color to a lighter one, maybe the same as the page's background? Right now it is a bit too heavy in my opinion

```suggestion
      baseColor="#e5e5e5"
```

wlechowicz: Agreed, the colors could use some adjustments, the grey is IMO too dark and looks too "rough" against the background. But I'm not a color expert so all I Can say is "it looks kinda rough".

PawelPazderski: the scelleton is really cool feature, maybe loose that gradient animation? I like this kind of stuff, like YouTube just some nice grey elements:
image

apps/web/app/(navigation)/budgets/[id]/BudgetsContent.tsx
```diff
@@ -0,0 +1,59 @@
...
+          accept: "*/*",
+          "Content-Type": "application/json",
+          Authorization: "Bearer " + token,
+        },
+      }).then((res) => res.json());
+    },
+  });
+  const mainCardContent = (
+    <BudgetContentWrapperStyled>
+      <BudgetBasicInformation budget={budget} />
```
Comment (zuzannapaw) [inline apps/web/app/(navigation)/budgets/[id]/BudgetsContent.tsx:45]: It looks great with skeleton!! ✨ My small question about this part: I saw that you placed `<BudgetBasicInformationSuspense />` inside `<BudgetBasicInformation/>`, and if there is no budget that came from props, the suspense is rendered.  Shouldn't it be the way that  `<BudgetBasicInformationSuspense />`  is in separated file, and here, in `const mainCardContent` there is a condition f.ex `budget?  <BudgetBasicInformation/> : <BudgetBasicInformationSuspense />` . What should be rendered would be defined directly in the components' tree and <BudgetBasicInformation/> would be kept neat. Let me know what you think! 💭 👀

packages/ui/SkeletonLoading/index.tsx
zuzannapaw: @PawelPazderski  I agree, just grey elements or maybe a softer gradient would look nice! 👆  Useful thing that comes with gradient is that it moves, so user knows that some data is in loading process and it will be served for him in a moment. ☺️

Alessiass (approved):

monika-klokosinska-intive (approved):

PR #80, opened 2023-05-08
fix: add missing link in login form #P23-304

zuzannapaw:
ticket: https://tracker.intive.com/jira/browse/PATRO23-304

- add missing text with link in log-in form to be able to move to sign-up page ➡️ 
- dictionnary has been updated! 📖 

Element that was missing: 
image

<h2>Update 📣 </h2>

Due to acheive proper styles from design and avoid code repetition,  component `<FormFooter>` has been created. It receives props such as: 

```typescript
type FormFooterProps = {
  basicText: string;
  linkText: string;
  href:string;
};
```  
In connection with this change, component was added to Storybook as well  📖

### DIFF
--- a/apps/web/app/(regflow)/sign-in/page.tsx
+++ b/apps/web/app/(regflow)/sign-in/page.tsx
@@ -3,7 +3,7 @@
 import { Field, Form } from "houseform";
 import { useState } from "react";
 import { useRouter } from "next/navigation";
-import { ErrorMessage, Button, Input, LinkComponent } from "ui";
+import { ErrorMessage, Button, Input, FormFooter } from "ui";
 import styled from "styled-components";
 import { z } from "zod";
 import { useTranslate } from "lib/hooks";
@@ -48,12 +48,6 @@ const ErrorSuportingMsg = styled.div`
   left: 10px;
   color: #b3261e;
   font-size: small;
-`;
-
-const FooterStyled = styled.div`
-  font-size: 16px;
-  line-height: 150%;
-  margin-top: 42px;
 `;
 
 export default function SignInPage() {
@@ -138,12 +132,11 @@ export default function SignInPage() {
             <Button onClick={submit} type="submit" fullWidth>
               {t(form.submitButton)}
             </Button>
-            <FooterStyled>
-              {t(form.footer)}{" "}
-              <LinkComponent href="/sign-up">
-                {t(form.footerLink)}
-              </LinkComponent>
-            </FooterStyled>
+            <FormFooter
+              basicText={t(form.footer)}
+              linkText={t(form.footerLink)}
+              href="/sign-up"
+            />
           </form>
         </FormWrapper>
       )}
--- a/apps/web/app/(regflow)/sign-in/page.tsx
+++ b/apps/web/app/(regflow)/sign-in/page.tsx
@@ -3,7 +3,7 @@
 import { Field, Form } from "houseform";
 import { useState } from "react";
 import { useRouter } from "next/navigation";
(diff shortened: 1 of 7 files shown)


apps/web/lib/dictionary.tsx
```diff
@@ -104,6 +104,16 @@ const dictionary = {
         },
       },
       submitButton: { en: "Log In", pl: "Zaloguj", fr: "Se connecter" },
+      footer: {
+        en: "Don't have an account?",
+        pl: "Nie posiadasz konta?",
+        fr: "N'avez-vous pas de compte ?",
```
wyola, line 110:
```suggestion
        fr: "N'avez-vous pas de compte?",
```

zuzannapaw: @wyola  In french before some characters ( f.ex "?" or "!" ) the space is necessary ☺️

wyola: note to myself: don't comment on french ever again! lol

apps/web/app/(regflow)/sign-in/page.tsx
```diff
@@ -50,6 +50,12 @@ const ErrorSuportingMsg = styled.div`
   font-size: small;
 `;
 
+const FooterStyled = styled.div`
+  font-size: 16px;
+  line-height: 150%;
+  margin-top: 42px;
+`;
```
wyola, line 57: I just noticed that text in footer is missing color, it should be `#3E4C59`. The same problem is on sign-up page (which sadly I was doing and missed it there🫥). Browser displays it as black I think. But I am actually wondering if we shouldn't have this added on `body`? 🤔

https://user-images.githubusercontent.com/80450315/236907837-0d175643-8841-47fe-830d-36e9f112ed2e.mp4

zuzannapaw: @wyola  great remark! I can add styled component for it with a proper color. What do you think?

apps/web/lib/dictionary.tsx
zuzannapaw: @wyola don't worry! French grammar can be a true nightmare! 👀

Vectari (approved):

apps/web/app/(regflow)/sign-in/page.tsx
zuzannapaw: @wyola Fixed!

packages/ui/FormFooter/index.tsx
```diff
@@ -0,0 +1,33 @@
+"use client";
+import { ReactElement } from "react";
```
wyola, line 2: this is not used

packages/ui/FormFooter/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import styled from "styled-components";
+import { LinkComponent } from "ui";
+
+type FormFooterProps = {
+  basicText: string;
+  linkText: string;
+  href:string;
```
wyola, line 9:
```suggestion
  href: string;
```

wyola (approved):
Just two little things, other than that it is good to go I think! ⭐

packages/ui/FormFooter/index.tsx
zuzannapaw: Thanks! Removed!

packages/ui/FormFooter/index.tsx
zuzannapaw: Fixed!

monika-klokosinska-intive (approved):

PR #81, opened 2023-05-08
feat: integrate create new budget backend #P23-291 

Vectari:
### Hello! Here is PR of my task [P23-291](https://tracker.intive.com/jira/browse/PATRO23-291).
1. Because for now working Currency from BE is only USD and EUR I left only this two in Currency Select list.
2. I use uuid to generate individual `id`.

## How to use:
1. Use `npm i`
2. Go to [Swagger](https://inbudget-patronage-api-dev.azurewebsites.net/swagger/index.html) and use `{
  "email": "smutnarzaba@png.pl",
  "password": "Frytki123.FE"
}` in `/user/sign-in`
3. Copy token to **_133_** line of my code and then `npm run dev`
4. After you create new budget in app and **Authorize in Swagger** go to `/budgets/list` and use `{
  "pageSize": 20,
  "pageIndex": 1,
  "search": "budget",
  "sortDescriptors": [
    {
      "columnName": "name",
      "sortAscending": true
    }
  ]
}
` to show budget list. Change `pageSize` if longer list is needed.
5. Copy `id` value and paste in `budgets/{id}` if you want check budget detail.

## Have a great day! 😁


apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -118,6 +121,55 @@ export const CreateNewBudget = ({ onClose }: NewBudget) => {
...
+
+  const queryClient = useQueryClient();
```
bdeluga, line 135: We are already creating one instance of queryClient in RootLayout, so this line is redundand.

apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -118,6 +121,55 @@ export const CreateNewBudget = ({ onClose }: NewBudget) => {
...
+            },
+            description: `${newBudget.description}`,
+            iconName: `${newBudget.icon}`,
+          }),
+        }),
+      {
+        onSuccess: () => {
+          queryClient.invalidateQueries("budgets");
+        },
+        onError: () => {},
```
bdeluga, line 167: Maybe, let's at least console.log if any error do happen. 😵

apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -118,6 +121,55 @@ export const CreateNewBudget = ({ onClose }: NewBudget) => {
     currentLang === "pl" && setLang("pl-PL");
   }, [lang, currentLang]);
 
+  // TRANSFORM TIMESTAMP TO ISO (+7200000 miliseconds, unix timestamp is in seconds, but Javascript counts in miliseconds, * 1000 not work)
+  const startDateTimestamp = newBudget.dateStart + 7200000;
+  const endDateTimeStamp = newBudget.dateEnd + 7200000;
+  const budgetStartDate = new Date(startDateTimestamp).toISOString();
+  const budgetEndDate = new Date(endDateTimeStamp).toISOString();
+
+  const url = "https://inbudget-patronage-api-dev.azurewebsites.net/budgets";
```
bdeluga, line 130: #69 is already on main, please use .env for that 📦

bdeluga (changes requested):
Small things to improve. 📖

apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -118,6 +121,55 @@ export const CreateNewBudget = ({ onClose }: NewBudget) => {
...
+              value: `${newBudget.limit}`,
+              currency: `${newBudget.currency.tag}`,
+            },
+            period: {
+              startDate: `${budgetStartDate}`,
+              endDate: `${budgetEndDate}`,
+            },
+            description: `${newBudget.description}`,
+            iconName: `${newBudget.icon}`,
+          }),
```
wyola, line 161: You are using string interpolation on every value here, while everything, except the value, is a string already. And value should be sent to the backend as a number, so also not needed here.

```suggestion
          body: JSON.stringify({
            id: uuidv1(),
            name: newBudget.name,
            userId: uuidv1(),
            limit: {
              value: newBudget.limit,
              currency: newBudget.currency.tag,
            },
            period: {
              startDate: budgetStartDate,
              endDate: budgetEndDate,
            },
            description: newBudget.description,
            iconName: newBudget.icon,
          }),
```

apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -1,6 +1,9 @@
 "use client";
 
 import { useContext, useEffect, useState } from "react";
+import { useMutation, useQueryClient } from "react-query";
+import { v1 as uuidv1 } from "uuid";
```
aniasgit, line 5: why did you choose timestamp uuid not a random one (v4)?

apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -118,6 +121,55 @@ export const CreateNewBudget = ({ onClose }: NewBudget) => {
     currentLang === "pl" && setLang("pl-PL");
   }, [lang, currentLang]);
 
+  // TRANSFORM TIMESTAMP TO ISO (+7200000 miliseconds, unix timestamp is in seconds, but Javascript counts in miliseconds, * 1000 not work)
+  const startDateTimestamp = newBudget.dateStart + 7200000;
+  const endDateTimeStamp = newBudget.dateEnd + 7200000;
+  const budgetStartDate = new Date(startDateTimestamp).toISOString();
+  const budgetEndDate = new Date(endDateTimeStamp).toISOString();
```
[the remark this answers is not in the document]
PawelPazderski: Also, I think we should use one date format same for every place, like we got day/month/year in the tile:
image
when in the date picker is month/day/year:
image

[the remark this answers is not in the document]
aniasgit: It should depend on locale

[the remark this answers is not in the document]
monika-klokosinska-intive: @aniasgit is right about the datetime manipulation. You absolutely don't want to manage timezones etc. by hand. That path leads straight to hell. It's fine to transform one date type into another (e.g. from timestamp to ISOString if that' what the backend wants), but dealing with timezones is best left to an existing JS utility / library.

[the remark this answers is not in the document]
Vectari: You are probably right, but i think it should be new task to fix/refactor datepicker. Here I just integrate Create New Budget with BE. Of course, I can do that if this task is the place for such correction. About that 2 hours. I did this in that way because without it when you pick for example 10 June budget send to BE 9 June, time 22:00. I know it’s wild but I was thinking if time is not important for us so I can make it in that way. I try other way than :)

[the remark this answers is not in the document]
AzulloCyll: > Also, I think we should use one date format same for every place, like we got day/month/year in the tile: image when in the date picker is month/day/year: image

You can change datepicker date format in CustomDatepicker component in ui. 
image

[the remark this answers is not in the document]
Vectari: Yes I know. I change it too then

apps/web/app/(navigation)/CreateNewBudget.tsx
Vectari: I think it's required for onSuccess but I will back to onSuccess and onError tomorrow and check.

apps/web/app/(navigation)/CreateNewBudget.tsx
Vectari: I add error tomorrow 🙃

apps/web/app/(navigation)/CreateNewBudget.tsx
Vectari: Done :)

apps/web/app/(navigation)/CreateNewBudget.tsx
Vectari: Done :)

apps/web/app/(navigation)/CreateNewBudget.tsx
Vectari: I changed to v4. Later uuid will be deleted because we dont have to generate and send ID but for now request without ID not work in BE.

apps/web/app/(navigation)/CreateNewBudget.tsx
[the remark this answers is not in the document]
Vectari: I deleted +2h and change Date Picker to dd/MM/yyyy. Hope everything is good now 😄

apps/web/app/(navigation)/CreateNewBudget.tsx
Noshi96: @Vectari Btw, you actually don't need to add external libs to generate v4 uuid. You can use the native JS function 🚀  https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

Vectari: You are right. I changed to this one 👍

apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -120,6 +124,53 @@ export const CreateNewBudget = ({ onClose }: NewBudget) => {
     currentLang === "pl" && setLang("pl-PL");
   }, [lang, currentLang]);
 
+  const { data: session } = useSession();
+
+  const startDateTimestamp = newBudget.dateStart;
+  const endDateTimeStamp = newBudget.dateEnd;
+  const budgetStartDate = new Date(startDateTimestamp).toISOString();
+  const budgetEndDate = new Date(endDateTimeStamp).toISOString();
```
aniasgit, line 132: It won't be needed if https://github.com/intive/patronage2023-js/pull/81/files#r1191108946

apps/web/app/(navigation)/CreateNewBudget.tsx
aniasgit: This queryClient will be needed for invalidate budgets.

bdeluga (approved):

apps/web/app/(navigation)/CreateNewBudget.tsx
```diff
@@ -50,8 +54,8 @@ type newBudgetType = {
   limit: number | string;
   description: string;
   icon: string;
-  dateStart: number | null;
-  dateEnd: number | null;
+  dateStart: any;
+  dateEnd: any;
```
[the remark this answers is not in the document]
Vectari: I will try but I can't make it before demo 😕

PR #82, opened 2023-05-09
feat: create new income and expense #P23-232

aniasgit:
task: [https://tracker.intive.com/jira/browse/PATRO23-232](url)

What is done:

- form for creating new transactions,
- category selector,
- integration with BE

To do:
-notifications about success or failure when adding a transaction


PawelPazderski (approved):
Looks and works very nice, good job! The only thing that bothers me while interacting is the jumping effect caused by disabling the scrollbar "the modal mode" when opening Select for a category, in Dropdown it was possible to set it as `modal={false}` but can't see that for Select which would be cool here as well. 
image

There is also an error in console:
image

wyola:
> Looks and works very nice, good job! The only thing that bothers me while interacting is the jumping effect caused by disabling the scrollbar "the modal mode" when opening Select for a category, in Dropdown it was possible to set it as `modal={false}` but can't see that for Select which would be cool here as well. image
> 
> There is also an error in console: image

@PawelPazderski there is a fix for the jumping effect on @Noshi96 's branch, it will be changed in global CSS 👉 https://github.com/intive/patronage2023-js/pull/71#discussion_r1187704186

PawelPazderski:
> > Looks and works very nice, good job! The only thing that bothers me while interacting is the jumping effect caused by disabling the scrollbar "the modal mode" when opening Select for a category, in Dropdown it was possible to set it as `modal={false}` but can't see that for Select which would be cool here as well. image
> > There is also an error in console: image
> 
> @PawelPazderski there is a fix for the jumping effect on @Noshi96 's branch, it will be changed in global CSS 👉 [#71 (comment)](https://github.com/intive/patronage2023-js/pull/71#discussion_r1187704186)

apps/web/app/(navigation)/budgets/CreateNewTransaction.tsx
```diff
@@ -0,0 +1,236 @@
...
+                    z
+                      .string()
+                      .nonempty({ message: t(dict.errors.amountNotEmpty) }),
+                    z.number().positive({
+                      message: t(dict.errors.amountGraterThanZero),
+                    }),
+                  ])}>
+                  {({ value, setValue, errors }) => (
+                    <Input
+                      name="amount"
```
wyola, line 154: "Amount" input does not focus when clicking on label. It is because `id` for it is created based on `name` and it is not unique within the page - ka-table creates id "amount" based on column's name. So you either need to change name here or add id prop with unique value.

apps/web/app/(navigation)/budgets/CategorySelector.tsx
```diff
@@ -0,0 +1,94 @@
...
+  SelectContentStyled,
+  SelectIconStyled,
+  SelectItemStyled,
+  SelectPortalStyled,
+  SelectTriggerStyled,
+  SelectTriggerWrapperStyled,
+  SupportingLabelStyled,
+} from "./CategorySelectorStyled";
+import { CategoryIcon, Icon } from "ui";
+import categoryMap from "../../../lib/category-map";
```
wyola, line 15:
```suggestion
import categoryMap from "lib/category-map";
```

wyola (commented):
Few things related mostly to code repetition in styles. Idk if maybe you were in a hurry and intend to change this later, I will just leave it here.

On general note, I think that `select` should be an atomic component and serve both for Currency and Categories

Noshi96:
> > Looks and works very nice, good job! The only thing that bothers me while interacting is the jumping effect caused by disabling the scrollbar "the modal mode" when opening Select for a category, in Dropdown it was possible to set it as `modal={false}` but can't see that for Select which would be cool here as well. image
> > There is also an error in console: image
> 
> @PawelPazderski there is a fix for the jumping effect on @Noshi96 's branch, it will be changed in global CSS 👉 [#71 (comment)](https://github.com/intive/patronage2023-js/pull/71#discussion_r1187704186)

Indeed, the solution was given, but it is slightly different. The problem was that the scrollbar from radix was forced to be removed when the option was opened. Now, there is a custom hook that checks if we have an active scrollbar on the page. Please add the lines below to `CategorySelector.tsx` and the problem goes away.  

```js
const { hasScrollbar } = useHasScrollBar();
```

```js
<SelectPortalStyled className={hasScrollbar ? "radix-scroll" : "">
  <SelectContentStyled position="popper">
    <Select.Viewport>
```

The magic happened here, in `global.css` where we force the scroll bar to be maintained when the options were opened, when class `radix-scroll` was included:

```css
body:has(.radix-scroll) {
  @media only screen and (min-width: 768px) {
      overflow-y: scroll !important;
  } 
}
```

apps/web/app/(navigation)/budgets/CategorySelector.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <SupportingLabelStyled hasError={hasErrors}>
+          {errors && errors[0]}
+        </SupportingLabelStyled>
+      </SelectTriggerWrapperStyled>
+
+      <SelectPortalStyled>
+        <SelectContentStyled position="popper">
+          <Select.Viewport>
+            {getCategoriesItems(categoryMap)}
+            {/* {categories.map((category) => (
```
monika-klokosinska-intive, line 84: @aniasgit if you don't need this piece of code anymore, you can remove it altogether.

apps/web/app/(navigation)/budgets/CategorySelectorStyled.tsx
```diff
@@ -0,0 +1,98 @@
+import * as Select from "@radix-ui/react-select";
+import styled, { css } from "styled-components";
+
+export const SelectTriggerWrapperStyled = styled.div`
+  position: relative;
+`;
+
+export const SelectTriggerStyled = styled(Select.Trigger)<{
```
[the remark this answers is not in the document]
monika-klokosinska-intive: @wyola one component for both might not work, since the behaviour is different, but it would be good to re-use thee styles in both places if possible.

apps/web/app/(navigation)/budgets/CreateNewTransactionStyled.tsx
```diff
@@ -0,0 +1,78 @@
...
+`;
+
+export const DatePickerErrorStyled = styled.div`
+  position: absolute;
+  color: ${({ theme }) => theme.textarea.error};
+  font-weight: 400;
+  font-size: 12px;
+  margin: 4px 10px 0 10px;
+  top: 55px;
+`;
```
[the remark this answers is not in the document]
aniasgit: I agree that a lot of styles are duplicated, but these are styles for separate forms. I don't want the style changes in createNewBudget to affect the appearance of my form.

apps/web/app/(navigation)/budgets/CreateNewTransaction.tsx
aniasgit: changed name

aniasgit:
> Few things related mostly to code repetition in styles. Idk if maybe you were in a hurry and intend to change this later, I will just leave it here.
> 
> On general note, I think that `select` should be an atomic component and serve both for Currency and Categories

I agree that select should be an atomic component. When I was creating it, I suggested Wojtek's statement that it is part of the form and I placed it next to the form. I think I could move it to the ui and let it take whatever components we want, but it requires some changes. I don't want to do it before the demo. However, it will not be the same select for currency and category, because in currency there is a bit of a custom display of the selected value (the tag itself)

packages/ui/theme.tsx
```diff
@@ -223,6 +223,14 @@ export const theme = {
     tag: colors.Neutral8,
     tagFocus: colors.Teal8,
   },
+  categorySelect: {
```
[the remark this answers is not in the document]
aniasgit: If we decided to do separate objects in theme form separate components I think we should stay with this method.

apps/web/app/(navigation)/budgets/CreateNewTransactionStyled.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const ParagraphStyled = styled.p`
+  font-size: 12px;
+  font-weight: 600;
+  margin-bottom: 24px;
+  padding-left: 3px;
+  padding-top: 56px;
+  color: #515151;
```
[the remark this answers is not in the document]
aniasgit: done

apps/web/app/(navigation)/budgets/CreateNewTransactionStyled.tsx
```diff
@@ -0,0 +1,78 @@
...
+`;
+
+export const ContentStyled = styled.div<{ fullHeight?: boolean }>`
+  display: grid;
+  overflow-y: auto;
+  grid-row-start: 1;
+  grid-row-end: 2;
```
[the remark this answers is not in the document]
aniasgit: removed

apps/web/app/(navigation)/budgets/CategorySelectorStyled.tsx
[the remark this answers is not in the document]
aniasgit: similar to this https://github.com/intive/patronage2023-js/pull/82#discussion_r1190229252

apps/web/app/(navigation)/budgets/CategorySelector.tsx
aniasgit: removed

wlechowicz:
Review authors, it is on you to resolve conversations that are done. @monika-klokosinska-intive @wyola and everyone.

bdeluga (approved):
If it builds, it can be merged 🚀

Noshi96 (approved):

apps/web/lib/dictionary.tsx
```diff
@@ -456,6 +456,94 @@ const dictionary = {
...
+    errors: {
+      min3characters: {
+        en: "Name must have at least 3 characters.",
+        pl: "Nazwa musi zawierać co najmniej 3 znaki.",
+        fr: "Le nom doit contenir au moins 3 caractères.",
+      },
+      max58characters: {
+        en: "Name must not have more than 58 characters.",
+        pl: "Nazwa nie może mieć więcej niż 58 znaków.",
+        fr: "Le nom ne peut pas dépasser 58 caractères.",
```
wlechowicz, line 518: If requirement changes to be 59 instead of 58 chars, it would be a lot of changes here and in the component. Instead, this should just be a function that takes the number and puts it into the error string `Name must not have more than ${length} characters.`

wlechowicz (changes requested):

apps/web/lib/dictionary.tsx
wlechowicz: As discussed, since this is done this way in multiple places, we should aim to improve this solution in a separate, refinement task.

apps/web/app/(navigation)/budgets/[id]/CreateNewTransaction.tsx
```diff
@@ -0,0 +1,246 @@
...
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(newTransaction),
+      });
+      console.log(result.then((Response) => Response.json()));
+      return result;
+    },
+    {
+      onSuccess: () => {
+        // queryClient.invalidateQueries(["datatable"]);
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/CreateNewTransaction.tsx:64]: @aniasgit we gonna add the invalidation back in!

monika-klokosinska-intive (changes requested):
@aniasgit since the main "meat" of this PR is already merged, please follow up with a refinement PR addressing the requested changes.

PR #83, opened 2023-05-10
feat: trend chip #P23-309

zuzannapaw:
ticket: https://tracker.intive.com/jira/browse/PATRO23-309

<h3>Overview 👀 </h3>
Trend chip component that accepts value in props. Depending on a positive or negative value, it takes proper color and icon.
 <h4>Update! 💥  </h4> Component takes as well `ariaLabel` prop.
 
```typescript
export type TrendChipProps = {
  value: number;
  ariaLabel: string;
```  

<h2> Update: problem resolved ✔️ </h2>
<h3>Problem 🤔</h3>

I tried to imagine a situation where user can get really high income, and then value could be quite long. In that situation, even though TrendChip doesn't have width, but only paddings, the value goes beyond it's border. Now this problem is caused by wrapper in storybook, so the width depends on parent element. Should it be that way, or maybe we need to find another solution?

<h3>Component's design: </h3>

image
image

### DIFF
--- a/packages/ui/TrendChip/index.tsx
+++ b/packages/ui/TrendChip/index.tsx
@@ -1,9 +1,11 @@
 "use client";
 import styled from "styled-components";
 import { Icon } from "ui";
+import { IconType } from "../Icon";
 
 export type TrendChipProps = {
   value: number;
+  ariaLabel?: string;
 } & React.HTMLProps<HTMLDivElement>;
 
 export const TrendChipStyled = styled.div<TrendChipProps>`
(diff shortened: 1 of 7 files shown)


bdeluga (approved):
LGTM! 🚀

monika-klokosinska-intive (changes requested):
@zuzannapaw You are missing the third use case mentioned in the ticket: gray colors and flat arrow (line if there is no arrow?) when the passed value equals zero. 
There is no design for this usecase so you have to use your creativity ;).

packages/ui/TrendChip/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const displayedValue =
+    value > 0 ? <span>{value}%</span> : <span>{Math.abs(value)}%</span>;
+
+  return (
+    <TrendChipStyled value={value}>
+      {displayedIcon}
+      {displayedValue}
+    </TrendChipStyled>
+  );
+};
```
Noshi96, line 46:
```suggestion
export const TrendChip = ({ value }: TrendChipProps) => {
  const iconName = (): IconType => {
    if (value === 0) return "trending_flat";
    if (value < 0) return "trending_down";
    return "trending_up";
  };
  return (
    <TrendChipStyled value={value}>
      <Icon icon={iconName()} iconSize={15} />
      <span>{Math.abs(value)}%</span>
    </TrendChipStyled>
  );
};
```

I think we can simplify it. What do you think?

packages/ui/TrendChip/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export const TrendChipStyled = styled.div<TrendChipProps>`
+  display: flex;
+  justify-content: center;
+  align-content: center;
+  gap: 4px;
+  line-height: 16px;
+  font-size: 12px;
+  padding: 2px 28px;
+  font-weight: 600;
```
Noshi96, line 17:
```suggestion
  font-size: 12px;
  padding: 4px 8px;
  font-weight: 600;
```

packages/ui/TrendChip/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export const TrendChip = ({ value }: TrendChipProps) => {
+  const displayedIcon =
+    value > 0 ? (
+      <Icon icon="trending_up" iconSize={15} />
+    ) : (
+      <Icon icon="trending_down" iconSize={15} />
+    );
+  const displayedValue =
+    value > 0 ? <span>{value}%</span> : <span>{Math.abs(value)}%</span>;
```
Noshi96, line 38: I wonder if we need some aria role status here, for example 🤔 https://stackoverflow.com/questions/68170589/do-html-elements-like-pill-badge-need-extra-aria-role

Noshi96 (approved):
With `width: fit-content` mentioned by @monika-klokosinska-intive will be perfect. Well done! 🚀  🚅

wlechowicz:
Doesn't look quite right in the storybook preview (vertical misalignment):
image

packages/ui/TrendChip/index.tsx
zuzannapaw: Great remark! On my way to apply it!

packages/ui/TrendChip/index.tsx
```diff
@@ -0,0 +1,46 @@
+"use client";
+import styled from "styled-components";
+import { Icon } from "ui";
+
+export type TrendChipProps = {
+  value: number;
+} & React.HTMLProps<HTMLDivElement>;
+
+export const TrendChipStyled = styled.div<TrendChipProps>`
```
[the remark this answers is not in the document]
zuzannapaw: Thank you so much! On my way to correct it!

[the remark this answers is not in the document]
zuzannapaw: @monika-klokosinska-intive I have applied your suggestions. I added new styles for the case when value is 0, cleaned up the code as @Noshi96 suggested, and added aria-label to the element. The thing is, we can't use dictionnary in /ui folder, therefore text for aria-label is being passed by props. In order to be able to pass to the aria label the proper value  that is being passed as well in prop `value` , I thought about this solution: 

image

It tried it and it seems to work:
   
image

With this change, I update the dictionary as well. What do you think about that?

packages/ui/TrendChip/index.tsx
zuzannapaw: @Noshi96 done! Thank you!

packages/ui/TrendChip/index.tsx
zuzannapaw: @Noshi96  done!

packages/ui/TrendChip/index.tsx
zuzannapaw: @Noshi96 I added it and i think it works fine!

zuzannapaw (commented):
@wlechowicz I fixed it and I hope it looks good now! 👀

PawelPazderski (approved):
Looks and works nice! Only thing I can see is that the "0" option i slighhtly smaller then the others, the icon is flat arrow so I guess it's the padding making the difference, overall great 🚀

monika-klokosinska-intive:
@wlechowicz this issue seems to be related to storybook not knowing how to load material-icons correctly. I get it sometimes as well, but when I refresh everything is looking good again.

monika-klokosinska-intive (approved):

PR #85, opened 2023-05-10
feat:  list of budgets - integration with be #P-23-34

AzulloCyll:
Hello. This is integration of budget list with BE.

Jira task: https://tracker.intive.com/jira/browse/PATRO23-34

What we have done? 
- changed display of left menu and "Add new budget button" to fixed position - now button is not rendered down off the visible part of screen
- fetched data from budget list endpoint and showed it on screen
- added sorting functionality
- added search functionality
- added infinite scroll

... and more little tweaks and fixes :)

thanks to @Noshi96 for helping me with that task!


wlechowicz:
Using Vercel preview, after I logged in, I saw a long list of budgets immediately in the menu. But after I refreshed the page, the list shows a spinner and "something went wrong".

Noshi96:
> Using Vercel preview, after I logged in, I saw a long list of budgets immediately in the menu. But after I refreshed the page, the list shows a spinner and "something went wrong". Then, if I switch to a different tab in the browser, then go back to inbudget, the list is there again. When I refresh, it breaks again.
> 
> Also, clicking budgets on the list changes the Id in the URL, but the budget details don't change. It always shows "aaaaaaa" budget even if I refresh the page with other budget's Id in URL.

Details of the budget didn't change because budget page [id] had hardcoded id 😅  It's already fixed.

The first problem was caused by sending a POST while the session token wasn't available. The behavior where you saw the refreshed data was caused by tanstaq Window Focus Refetching https://tanstack.com/query/v4/docs/react/guides/window-focus-refetching. So after the refocus you already had the token and the data was displayed. It's also fixed. 🚀

apps/web/lib/hooks/useGetBudgets.tsx
```diff
@@ -0,0 +1,33 @@
...
+  sortAscending: boolean,
+  pageSize: number
+) => {
+  const { data: sessionData } = useSession();
+  const axiosInstance = reqInstance(sessionData?.user.accessToken);
+
+  return useInfiniteQuery({
+    queryKey: ["budgets", { searchValue, sortAscending }],
+    queryFn: async ({ pageParam = 1 }) => {
+      await sleep(600);
```
monika-klokosinska-intive, line 19: @AzulloCyll please remove this!

apps/web/lib/hooks/useGetBudgets.tsx
```diff
@@ -0,0 +1,33 @@
...
+        pageParam,
+        searchValue,
+        sortAscending,
+        axiosInstance,
+      });
+    },
+    getNextPageParam: (lastPage, allPages) => {
+      return lastPage.items.length ? allPages.length + 1 : undefined;
+    },
+    enabled: !!sessionData,
```
monika-klokosinska-intive, line 31: @AzulloCyll won't it work without the hack?

apps/web/lib/hooks/useGetBudgets.tsx
Noshi96: When there is no throttling, we don't have a nice looking skeleton loading animation, it pops up and shows for something around 100ms. In the video, the first part is without this throttling, and the second with this, await sleep().  @monika-klokosinska-intive do we want to remove it, though? 

https://github.com/intive/patronage2023-js/assets/38572172/7ddcc548-8c34-4ffa-ac16-d1994b8f121f

monika-klokosinska-intive: @Noshi96 it occured to me that you probably added it for testing / presenting purposes. I am fine with leaving it here, but please inform Ola and Bartek Deluga about this feature, so they can mention it during the presentation (e.g. budgets load is slowed in code on purpose, observe the beautiful skeleton loading dear clients).

wlechowicz: Let's have it in for demo and let's mention it explicitly like Monika said.

apps/web/services/mutations.ts
```diff
@@ -0,0 +1,54 @@
...
+    pageIndex: pageParam,
+    search: searchValue,
+    sortDescriptors: [
+      {
+        columnName: "name",
+        sortAscending: sortAscending,
+      },
+    ],
+  };
+  return (await axiosInstance.post<ItemType>("/budgets/list", body)).data;
```
wlechowicz, line 53: When did we agree to use axios?

packages/ui/NavList/NavItem.tsx
```diff
@@ -50,21 +52,36 @@ export const NavItemStyled = styled(Link)<NavItemPropsTransient>`
...
-    <NavItemStyled $active={active} href={href} onClick={onClick}>
-      <ChildrenWrapper>{children}</ChildrenWrapper>
-      {active && <Icon icon="chevron_right" color="#1E4C40" iconSize={18} />}
-    </NavItemStyled>
-  );
-};
+export const NavItem = React.forwardRef<HTMLLIElement, NavItemProps>(
+  ({ active, href, children, onClick }, ref) => {
+    return (
+      <li ref={ref}>
```
wlechowicz, line 71: What is this extra node for?

apps/web/lib/hooks/useGetBudgets.tsx
Noshi96: Good point! I will inform them :D

packages/ui/NavList/index.tsx
```diff
@@ -1,26 +1,48 @@
...
+  error?: ReactNode;
+  text?: { noData?: string; loading?: string; error?: string };
   onNavListItemClick: () => void;
 } & React.HTMLProps<HTMLUListElement>;
 
 const NavListStyled = styled.ul`
   list-style: none;
   width: 100%;
-  height: auto;
+  height: calc(100vh - 310px);
```
wlechowicz, line 32: Oddly specific, why -310px?

apps/web/app/NavListDemo.tsx
```diff
@@ -20,6 +20,7 @@ export const dummyNavItemContents: Array<NavItemContents> = [
     ),
     href: "/bills",
     id: 1,
+    ref: () => {},
```
wlechowicz, line 23: This, being added to every item, looks like unnecessary repetition.

wlechowicz (approved):
I'm leaving comments that are change requests, but we will address them as a fast follow up, approving as is for demo and we will refactor right away.

bdeluga (approved):
Let's just merge so we can move forward demo preparations

apps/web/lib/hooks/useGetBudgets.tsx
AzulloCyll: removed

apps/web/services/mutations.ts
AzulloCyll: axios is not used anymore

packages/ui/NavList/index.tsx
AzulloCyll: It's from here:
untitled-1_720
Distance from button to end of the list was unspecified, so i decided to set it to 310px.

packages/ui/NavList/NavItem.tsx
AzulloCyll: Our NavList is <ul> so in lists we decided to add <li> elements because we want our html to be semanticaly correct.

apps/web/app/NavListDemo.tsx
AzulloCyll: That file was deleted, it was mocked data.

PR #86, opened 2023-05-10
feat: trend chart #P23-310

wyola:
## UPDATE
merged into https://github.com/intive/patronage2023-js/pull/88

-----------------------------------------

PR for a trend chart
task: https://tracker.intive.com/jira/browse/PATRO23-310

Notes:
- still working on colors - I did not figure out yet how to change line color for negative values and how to add gradient for fill
- I did not add tooltip and I don't know if it will make sense here tbh 👀 I am not sure, but I think that tooltip is displayed inside of the chart's `<canvas>` tag (will check it tomorrow) and this area is rather small. We discussed the tooltip during the meeting but I don't see it in the task @wlechowicz let me know what you think about it
- remember that now there is a big gap between the total balance and the chart - it will look better once it is put together with the range selector ;-)


PawelPazderski (commented):
Looks very good so far. I woludn't bother with the gap between - simply would put it in a certain position relative to totalbalance, no matter what the width is - something like info tiles, they stand to left leaving empty space on right, so IMO should your elements and what comes next to them do. 
Also I'm not sure if I get the design right - but I wolud leave the whole chart highlighted green when the trendchip is positive, and appropriately redish if negative, on the other hand, this way you show it shows where the 0 line is which is also good.. just thinking out loud.

wyola:
> Looks very good so far. I woludn't bother with the gap between - simply would put it in a certain position relative to totalbalance, no matter what the width is - something like info tiles, they stand to left leaving empty space on right, so IMO should your elements and what comes next to them do. Also I'm not sure if I get the design right - but I wolud leave the whole chart highlighted green when the trendchip is positive, and appropriately redish if negative, on the other hand, this way you show it shows where the 0 line is which is also good.. just thinking out loud.

@PawelPazderski from what I understood, the chart is created from datapoints which are:
- x axis - date
- y axis - balance for the date
- 
so I think it can happen that balance for the day is negative yet trend value for the whole budget is positive? There was a problem with backend yesterday and I could not add more data to see how statistics provided by them actually work. I will revisit this today and adjust if needed.

as for the layout, maybe you are right, but we will see once this is put together with what @bdeluga is doing :)

PawelPazderski:
> 

Yeah, sure the single balance can be negative for the day that's clear, just looking at the chart in design I thought that the idea is the chart shows the whole trend by its color, not showing whether the single points are over or below zero thats all. 
### Honestly I prefer your way that shows where 0 line actually is.

apps/web/app/(navigation)/budgets/BudgetsContent.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+`;
+
+export const BudgetsContent = () => {
+  const id = usePathname()?.replace("/budgets/", "");
+
+  const [budgets, setBudgets] = useState<Budget[]>([]);
+
+  useEffect(() => {
+    fetch(`/budgets.json`)
```
monika-klokosinska-intive, line 34: @wyola I take all of this is now unnecessary, since we you will get the data from parent, right?

apps/web/app/(navigation)/budgets/[id]/TrendChart.styled.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  ${device.tablet} {
+    flex-direction: row;
+    justify-content: space-between;
+    gap: 6%;
+
+    > * {
+      flex: 1;
+      flex-shrink: 0;
+      width: 47% !important; //override inline styles added to <canvas>
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/TrendChart.styled.tsx:36]: I spy with my little eye an `!important`. Check if you can increase the css specifically for this rule, so the style `wins out` without using `!important`.

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
```diff
@@ -0,0 +1,185 @@
...
+function getGradient(
+  ctx: CanvasRenderingContext2D,
+  chartArea: ChartArea,
+  min: number,
+  max: number
+) {
+  const chartWidth = chartArea.right - chartArea.left;
+  const chartHeight = chartArea.bottom - chartArea.top;
+
+  const green = "#92CE78";
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:59]: @wyola you probably want to move these to theme, get them from theme and pass them as function arguments here.

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
```diff
@@ -0,0 +1,185 @@
...
+    const bTimestap = new Date(b.datePoint).getTime();
+    return aTimestap - bTimestap;
+  });
+
+  const dates = [];
+  const values = [];
+  for (let item of sortedData) {
+    dates.push(item.datePoint);
+    values.push(item.value);
+  }
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:112]: @wyola avoid push if you can get the result using map. 
```
const dates = sortedData.map(data => data.datePoint);
const values = sortedData.map(data => data.value);
```

`map` is something we call a `clean function`, it has no side effects and given the same input data it will always return the same output.  `Push` however is not clean - the result will be different depending on the start value `dates` and `values`. Clean functions are easier to test and debug.

monika-klokosinska-intive (changes requested):
@wyola some fast forward change would be nice here ;).

monika-klokosinska-intive:
@wyola of course you don't have to work with this PR, since it's outdated anyway. Open a new one from main and address the fast-follow changes in it. Feel free to create a task in JIRA for this work.

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:59]: @monika-klokosinska-intive I will, I was adding this shortly before the demo and did not want to waste time on it

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
Comment (monika-klokosinska-intive) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:112]: @wyola this is a `it depends` situation. If the `sortedData` was a huge array, (million indexes), then your solution would be the way to go. Since we are operating relatively small-scale, readability and  security (less bugs, easier to debug) can take precedence over minuscule performance wins. But it is always a trade-off, so it's good to keep the pros and cons in mind.

apps/web/app/(navigation)/budgets/[id]/TrendChart.styled.tsx
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.styled.tsx:36]: @monika-klokosinska-intive library adds it as an inline style, so I don't know if there is anything that I can do about it 👀

wyola:
> @wyola of course you don't have to work with this PR, since it's outdated anyway. Open a new one from main and address the fast-follow changes in it. Feel free to create a task in JIRA for this work.

will do that!

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
Comment (monika-klokosinska-intive) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:59]: @monika-klokosinska-intive yikes, gonna go for `!important` then. Cases like this (overwriting third-party libraries stubborn styles) is the only acceptable place for `!important`. Sometimes you have to do what you have to do.

PR #87, opened 2023-05-10
feat: transaction table controller #P23-289

Alessiass:
### Overview 🚀 
It retrieves data from the server based on the information provided from the budget and utilizes the pagination component.

## What's new ✨ 
- Based on the JWT token from Next-Auth, user input on pagination and budget info the data is fetched from BE and saturates the Transaction Table component.
- Temporary spinner for table loading is added.
- In case of fetching failure an error window is displayed to the user with error code and after closing the window next fetching attempt is conducted.

## What is missing 👓 
- Skeleton loading has to be added later on.
- The budget ID passed from BudgetContent component is hardcoded at this moment.


apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const [currentPage, setCurrentPage] = useState<number>(1);
+  const [itemsPerPage, setItemsPerPage] = useState<number>(5);
+  const [totalPages, setTotalPages] = useState<number>(1);
+  const { t, dict } = useTranslate("BudgetsPage");
+  const setSorting = (column: string) => console.log(column);
+  const { data: session } = useSession();
+
+  const fixFetchedData = (res: APIResponse) => {
+    const tempArray = [] as Transaction[];
+    res.items.map((item) => {
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:41]: @Alessiass why not use regular map here?

Comment (Alessiass) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:42]: @monika-klokosinska-intive I'm note sure if I understand it right :D You mean to skip the temporary array and just return new array directly?

Comment (monika-klokosinska-intive) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:42]: @Alessiass that's why I meant! Doing a push inside a map is not a good look. Did you run into any async issues that forced you to use this approach?

Alessiass:
Branch already merged before demo

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (Alessiass) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:42]: I closed this pull request since it's already merged.
I've prepared that while waiting for pagination, next-auth and budget. The assumption was that we will get more data from budget prop and there will be need for additional mutations.
You are right and this is completly unecessary -> There is new pull request with fix.
Thanks for your input 🎉

PR #92, opened 2023-05-15
feat: administration panel #P23-317

bdeluga:
### Overview 📖 
[Chore](https://tracker.intive.com/jira/browse/PATRO23-317) was to implement administration panel.

## Features ✨ 
- Now we are able to get signed user role:
```tsx
//SomeFile.tsx

const { data:session } = useSession();

console.log(session?.user.role) // "ADMIN" || "USER"
```
- Users menu is dynamically rendered based on `user's role`

## Fixes 🔨 
- Fixed a lot of console warnings happening due to declaring `styled components` inside of component.
- Fixed logic for rendering active link's state of Navbar.

ups
- Fixed styles of NavbarLinkItem not taking whole width of parent


packages/ui/SideNavigationBar/index.tsx
```diff
@@ -1,5 +1,5 @@
 import styled from "styled-components";
-import { useState, ReactNode, useEffect, useMemo } from "react";
+import React, { useState, ReactNode, useMemo } from "react";
```
wyola, line 2: we don't need to import `React` and it is not used here

apps/web/app/(navigation)/SideNavigationBar.tsx
```diff
@@ -136,31 +139,50 @@ export default function SideNav() {
     ),
   };
 
+  //edit navbarItems below
+  const NavbarItems = [
```
wyola, line 143: it is just a variable, I don't think it needs capital letter in front

```suggestion
  const navbarItems = [
```

apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -12,23 +12,29 @@ type decodedData = {
...
   sub: string;
+  realm_access: {
+    roles: string[];
+  };
 };
 
 interface DefaultUser extends User {
   accessToken: string;
   refreshToken: string;
+  role: string;
```
Comment (wyola) [inline apps/web/pages/api/auth/[...nextauth].ts:23]:
```
  role: 'ADMIN' | 'USER';
```

or 
```
enum UserRole {
  Admin = "ADMIN",
  User = "USER",
}
...

role: UserRole,


// example usage: session?.user.role === UserRole.Admin
```

You could `export` the type and use it as well in `SideNavigationBar.tsx`. This way we don't have hardcoded `magic numbers` 🪄

wyola (commented):
Great job! Few small suggestions 🤗

PS I don't know within whose task is this, yours or @PawelPazderski but I think this title could be added in this PR, since Paweł's task is strictly about the table, and you create the page here.

image

apps/web/app/(navigation)/SideNavigationBar.tsx
```diff
@@ -136,31 +139,50 @@ export default function SideNav() {
...
+      icon: <Icon icon="settings" iconSize={32} />,
+      textValue: t(SideNav.settingsItem.title),
+      subMenu: settingsSubMenuData,
+      id: "3",
+    },
+  ];
+
+  const renderNavbar = () => {
+    //add items for admin
+    if (session?.user.role === "ADMIN") {
```
wlechowicz, line 168: As a non-admin user, I can easily get access to this by modifying state via dev tools. See if there is a way (there might not be, because we're forced into client components) to have this added server-side so the logic isn't easily accessible client-side.

apps/web/app/(navigation)/users/page.tsx
```diff
@@ -0,0 +1,10 @@
+import { DummyAsideCardContent } from "app/DummyAsideCardContent";
+import MultiCardLayout from "../MultiCardLayout";
+
+export default function ReportsPage() {
```
wlechowicz, line 4: Admin-only pages should be locked behind a middleware check to not allow non-admin users access to the pages, even if they "hack" their way around the menu. Also, component name.

```suggestion
export default function UsersPage() {
```

apps/web/app/(navigation)/users/page.tsx
```diff
@@ -0,0 +1,10 @@
...
+  const mainCardContent = <h1>Users</h1>;
+
+  return (
+    <MultiCardLayout main={mainCardContent} aside={<DummyAsideCardContent />} />
```
wlechowicz, line 8: The aside content for this page is dynamic, initially it's not there, then it shows user details once a user is selected. So this isn't very suitable.

wlechowicz (changes requested):

apps/web/app/(navigation)/users/page.tsx
bdeluga: It was hard to tell what was in scope of my task, since it had no description at all, so i just created blank page for @PawelPazderski  to work with.

wlechowicz: Too bad the description is empty, I agree. But the route and its security is definitely this task - adding the panel. Paweł's task is to fill it with users table (aka content).

apps/web/app/(navigation)/SideNavigationBar.tsx
bdeluga: I don't think there is a way to do this, as much as I would like to. By what i am testing every component inside of client component is also a client component (?). Omitting "use client" doesn't cause any errors, because component is already client-sided. 
image

image

Our layout is "use client" already which makes all of his children "use client" by default. So there is no way of using getServerSession and there is no way of actually rendering navbar based on user's role. 

I hope I am wrong and we can do something about it, I will think about it in the morning.

Moreover, you can easily change role, in react dev tools, but middleware will stop any kinds of breakthrough. 

image

bdeluga: Added middleware, it safely guards all sensitive routes.

mpiwonski (approved):
Good job @bdeluga!

wlechowicz (approved):
Obligatory 🚀 and LGTM

wyola (approved):
:shipit:

monika-klokosinska-intive (approved):

PR #93, opened 2023-05-15
refactor: trend chart #P23-330 

wyola:
## Trend chart refactor

Changes after last Thursday's battle 👀 
- added changes suggested in original PR https://github.com/intive/patronage2023-js/pull/86
- fixed error found during the demo (value of `addColorStop` going out of bounds)
- added theme and dictionary
- refactored some styles in parent component so the items are aligned properly

## New
I added "chart placeholder". Right now, when budget has less than 2 transactions, there is a white "hole" for the chart. I thought it would be nice to add there some placeholder. Let me know what you think!

image

⬇️ 

image

## Addition
We had wrong background color on budget's page 👀 I updated it to a correct one and added `theme` in related files.


apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
```diff
@@ -45,78 +46,147 @@ type TrendChartProps = {
...
+      gradient = ctx.createLinearGradient(
+        0,
+        chartArea.bottom,
+        0,
+        chartArea.top
+      );
+
+      gradient.addColorStop(0, negativeValueColor);
+      gradient.addColorStop(
+        Math.max(0, zeroPoint - 0.0000001),
```
Comment (wlechowicz) [inline apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:114]: Explain the decimal, please?

apps/web/lib/dictionary.tsx
```diff
@@ -601,6 +601,11 @@ const dictionary = {
         pl: "Bilans",
         fr: "Solde total",
       },
+      trendChartPlaceholder: {
+        en: "Add transactions to render chart",
+        pl: "Dodaj transakcje żeby wyświetlić wykres",
+        fr: "Ajoutez un transaction pour afficher le diagramme",
```
wlechowicz, line 607: The wording feels a bit awkward here, but I don't really have a good idea what the message should be. Team, get your suggestions in!

wlechowicz (changes requested):

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
```diff
@@ -45,78 +46,147 @@ type TrendChartProps = {
...
-  const dates = [];
-  const values = [];
-  for (let item of sortedData) {
-    dates.push(item.datePoint);
-    values.push(item.value);
+  const dates = sortedData.map((data) => data.datePoint);
+  const values = sortedData.map((data) => data.value);
+
+  const minValue = Math.min(...values);
+  const maxValue = Math.max(...values);
```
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:69]: @wlechowicz  ok, I've had the loop previously for the `dates` and `values` arrays. See here https://github.com/intive/patronage2023-js/pull/86#discussion_r1192234046

Would you suggest using `reduce` only for min and max values or maybe using one for loop for `dates`, `values`, `minValue` and `maxValue`? In such case `reduce` would not be needed. I have never used it before, it looks to me like it would make the code less readable 👀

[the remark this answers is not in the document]
Comment (wlechowicz) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:69]: Maybe it's less readable, but it's fun! 🤓 Just as an exercise. And you need to make sure you keep your job because nobody really understands parts of your code. Just kidding. Or am I.

```
  const { dates, values, minValue, maxValue } = sortedData.reduce((result, { value, datePoint }) => {
    // result is the accumulator
    result.dates = [ ...result.dates, datePoint ]; // Array.push is likely better here, but spread is more hip
    result.values = [ ...result.values, value ];
    if (value < result.minValue) {
      result.minValue = value;
    }
    if (value > result.maxValue) {
      result.maxValue = value;
    }

    return result;  
  }, { dates: [], values: [], minValue: 0, maxValue: 0 });
```

More code that is less readable, but only one pass over the array. Not worth it if you have 20 transactions, would be something to consider if you had 100,000 transactions.

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:114]: @wlechowicz I used gradient to color the line based on value. I followed the example from here https://www.chartjs.org/docs/latest/samples/advanced/linear-gradient.html

So the general idea is that line has different colors depending on value, but the gradient also has the moment where colors mix and I needed this to be a sharp "line"/border between the colors in the point where `value = 0`. 

If I used a value like `0.1` there would be more smooth change of the colors:
image
so I went for substracting a very small number, so that the difference is minimal and the border is sharp.

It was a bit in a hurry, so I just typed a lot of zeroes I admit 😅 now thinking about it, I could maybe use this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_VALUE ?

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:69]: @wlechowicz ok, will do that!

apps/web/lib/dictionary.tsx
wyola: @wlechowicz I was also wondering about this one, especially that you need at least two transactions to draw the chart.

"Add at least two transactions to see a chart"? seems a bit heavy as well 😅 idk, let's see what others think!

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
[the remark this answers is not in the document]
Comment (wlechowicz) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:69]: No, don't, it's pointless here. But maybe you can merge the 2 `map`s into single one, that might be worthwhile?

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:114]: @wlechowicz second parameter should be color and it works well if I do 
`gradient.addColorStop(zeroPoint, negativeValueColor);`
so I will simplify it this way, I overkilled a bit 🫥

apps/web/lib/dictionary.tsx
mpiwonski: Maybe we should form a sentence in another way like: "Chart will be displayed after adding at least 2 transactions". What do you think?

Vectari: I like @mpiwonski suggestion 👍

wyola: @wlechowicz @mpiwonski @Vectari sorry for the mix up, but I slept on it and came up with the idea to generate an SVG path for it to make it look like an image placeholder. Client accepted it so I will go in this direction! https://intive-apu8934.slack.com/archives/C04R6LM4JSV/p1684224052570909

Thanks anyway for your input!

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:69]: @wlechowicz @monika-klokosinska-intive could you please clarify what do you expect here, because I am lost right now 😅 how do I merge 2 maps into one? I need dates and values to be separate arrays to pass them to the chart. @monika-klokosinska-intive from what @wlechowicz wrote I understood that `reduce` is out of the game

apps/web/app/(navigation)/budgets/[id]/TrendChart.styled.tsx
```diff
@@ -1,25 +1,41 @@
...
+    margin-bottom: 16px;
+  }
 `;
 
 export const StyledBalanceChartWrapper = styled.div`
   display: flex;
   flex-direction: column;
   gap: 10px;
-  margin-bottom: 10px;
+  /* margin-bottom: 10px; */
```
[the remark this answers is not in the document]
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.styled.tsx:38]: I could not reach it from github, but shouldn't the gap value be in `px`? 

```css
 @media screen and (min-width: 1200px) {
    flex-direction: row;
    justify-content: space-between;
    gap: 6%;
```

apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx
```diff
@@ -1,6 +1,6 @@
 "use client";
 import { type BudgetGeneralInfo } from "lib/types";
-import React, { useMemo, useState } from "react";
+import React, { useState } from "react";
 import { TrendChip } from "ui";
```
Comment (zuzannapaw) [inline apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx:4]: `TrendChip` is not used in BudgetStatistics.tsx 👀 we could remove it from this file as well

zuzannapaw (commented):
Just a few comments from me 📝

apps/web/app/(navigation)/budgets/[id]/TrendChart.styled.tsx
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.styled.tsx:38]: @zuzannapaw it can be many things 🥸 I used `%` here, because it worked better when resizing the screen. I played with `px` and it never seemed to work good enough for different sizes

image

apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx]: fixed

apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx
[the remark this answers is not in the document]
Comment (wlechowicz) [reply apps/web/app/(navigation)/budgets/[id]/TrendChart.tsx:69]: Fine, let it be as it is now. I'll find another chance to torment you with Array methods.

wlechowicz (approved):

zuzannapaw (approved):

PR #94, opened 2023-05-15
fix: transactions table skeleton refinement #P23-320

zuzannapaw:
<h1>Transactions table skeleton refinement 🚧  </h1>

<h2>What changed ✏️ </h2>

- `isLoading` from `react query` has been passed to `<TransactionsTable/>`
- When the data is loading, all `th` are visible, they are not suspended. Skeleton is applied only inside `table body`
- Because we are not sure how many transactions will come to user, `gradient` has been applied as well
- By preserving the structure of html table, content is being properly displayed

<h2>Adjustement 🚀 </h2>

- By extending `ITableBodyProps` : 

```typescript
interface TransactionsTableSuspenseProps extends ITableBodyProps {
  rowsNumber: number;
}
``` 
it is possible now to pass prop `rowsNumber`  to `<TransactionsTableSuspense/>` . It defines how many `<DataRows/>` we would like to display.  Prop `rowsNumber` is used in function `spawnRows()` : 

```javascript
//function for creating array full of <DataRow/> components
const spawnRows = (props: TransactionsTableSuspenseProps) => {
  const rowsArray = [];
  let id = 1;

  for (let i = 0; i < props.rowsNumber; i++) {
    rowsArray.push(<DataRow {...props} key={id++} />);
  }

  return rowsArray;
};
``` 
Thanks to that, we are not limited in rows number.  If we would like to, we could pass here as well `itemsPerPage` to display the accurate number of rows that has been chosen by user in `Pagination`. 

- With classes from `/css/transactionsTableSuspense.css` ,the table is responsive 💻 

<h2> Update 💥  </h2>
Due to @Noshi96 suggestion:

> This makes me wonder if it's OK that the CSS directory is inside the /app. 🤔 If it were in the /web, we would simply write import "css/transactionsTableSuspense.css";

now `/css` folder is inside `/web` . Global styles' import has been changed as well in the main layout.


apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -110,7 +105,10 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
       <TransactionsTable
         currency={budget.currency}
         setSorting={setSorting}
-        transactions={transactionsData}
+        transactions={
+          transactionsData ? transactionsData : ([] as Transaction[])
```
Comment (wlechowicz) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:109]: It would be better if TransactionTable handled this, if the vale is not passed it should default to []. Also this could be simplified to `transactions={ transactionsData || [] }` but it's better to handle this via parameter default value in the component.

wlechowicz (changes requested):
LGTM, one improvement requested.

mpiwonski (approved):
Good job @zuzannapaw!

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:109]: @wlechowicz  Thanks for your comment, I will fix it tomorrow!

apps/web/app/(navigation)/budgets/[id]/TransactionsTableSuspense.tsx
```diff
@@ -1,91 +1,65 @@
 import { SkeletonLoading } from "ui";
 import styled from "styled-components";
 import { PropsWithChildren } from "react";
+import { ITableBodyProps } from "ka-table/props";
+import "../../../css/transactionsTableSuspense.css";
```
Comment (Noshi96) [inline apps/web/app/(navigation)/budgets/[id]/TransactionsTableSuspense.tsx:5]: This makes me wonder if it's OK that the CSS directory is inside the /app. 🤔 If it were in the /web, we would simply write import `"css/transactionsTableSuspense.css";`

apps/web/app/(navigation)/budgets/[id]/TransactionsTableSuspense.tsx
```diff
@@ -1,91 +1,65 @@
...
-          flexBasis: "15%",
-        }}>
-        {children}
-      </div>
-    );
-  };
+//to create single tr (thanks to tr, rows are properly displayed in table body)
+const DataRow: React.FC<ITableBodyProps> = ({ columns }) => {
+  return (
+    <tr>
```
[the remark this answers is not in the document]
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/TransactionsTableSuspense.tsx:29]: @Noshi96 when I didn't add that, typescript was screaming :/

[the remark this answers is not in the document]
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/TransactionsTableSuspense.tsx:29]: @Noshi96  thank you, i am going to apply your suggestion ☺️

apps/web/app/(navigation)/budgets/[id]/TransactionsTableSuspense.tsx
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/TransactionsTableSuspense.tsx:5]: @Noshi96  I tried that and I'm getting an error that this file is not found :/

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx]: @wlechowicz  it's fixed :)

apps/web/app/(navigation)/budgets/[id]/TransactionsTableSuspense.tsx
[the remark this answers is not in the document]
Comment (zuzannapaw) [reply apps/web/app/(navigation)/budgets/[id]/TransactionsTableSuspense.tsx]: @Noshi96  done! :)

Noshi96 (approved):
LGTM 🎸

wlechowicz (approved):

PR #96, opened 2023-05-16
feat: superfetch 🚀 

bdeluga:
### Overview 📖 
Extra chore to create superfetch to save us from code repetition madness 💫 

## Usage ✏️ 
All we have to do after importing useSuperfetch() is to initiate hook with desired url and options.

```tsx
//BudgetStatistics.tsx

//GET
  const fetch = useSuperfetch(
    `${env.NEXT_PUBLIC_API_URL}budgets/${id}/statistics?startDate=${startRange}&endDate=${endRange}`
  );
```
and then simply call our fetch

```tsx
const { data: statistics, isLoading } = useQuery({
    queryKey: ["rangedStatistics", startRange, endRange, budget.id],
    queryFn: async () => {
      return fetch().catch((err) => console.error(err));
    },
  });
```

### Another example

```tsx
//TransactionTableController.tsx

//POST

  const fetch = useSuperfetch(
    `${env.NEXT_PUBLIC_API_URL}budgets/${budget.id}/transactions`,
    {
      method: "POST",
      //no more JSON.stringify
      body: {
        pageSize: itemsPerPage,
        pageIndex: currentPage,
      },
    }
  );
```

then just fetch
```tsx

  const {
    data: transactionsData,
    isError,
    isLoading,
    refetch,
    error,
  } = useQuery({
    queryKey: ["datatable", itemsPerPage, currentPage, budget, session],
    queryFn: async () => {
      return fetch()
        .then((json) => fixFetchedData(json))
        .catch((err) => console.error(err));
    },
  });

```

## Look at this code save ✨ much wow...

From this:
```tsx

 const { data: statistics, isLoading } = useQuery({
    queryKey: ["rangedStatistics", startRange, endRange, budget.id],
    queryFn: async () => {
      return fetch(
        `${env.NEXT_PUBLIC_API_URL}budgets/${id}/statistics?startDate=${startRange}&endDate=${endRange}`,
        {
          headers: {
            accept: "*/*",
            "Content-Type": "application/json",
            Authorization: "Bearer " + session?.user.accessToken,
          },
        }
      ).then((res) => res.json());
    },
    enabled: !!session,
  });
```

To this:
```tsx

  const fetch = useSuperfetch(
    `${env.NEXT_PUBLIC_API_URL}budgets/${id}/statistics?startDate=${startRange}&endDate=${endRange}`
  );

  const { data: statistics, isLoading } = useQuery({
    queryKey: ["rangedStatistics", startRange, endRange, budget.id],
    queryFn: async () => {
      return fetch().catch((err) => console.error(err));

```

## Miscellaneous 🔖 
```ts
interface SuperOptions extends Omit<Omit<RequestInit, "headers">, "body"> {
  body: object;
}
```
Line above disable TS from giving us hints about headers object in fetch options and allows us to pass body as `object`

### DIFF
--- a/apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx
@@ -14,6 +14,7 @@ import {
 } from "./BudgetDetails.styled";
 import { StyledTrendChip } from "./TrendChart.styled";
 import { useTranslate } from "lib/hooks";
+import useSuperfetch from "lib/hooks/useSuperfetch";
 
 interface Props {
   budget: BudgetGeneralInfo;
@@ -59,23 +60,15 @@ const BudgetStatistics = ({ budget }: Props) => {
 
   const [startRange, endRange] = getRange();
 
-  const { data: session } = useSession();
+  const fetch = useSuperfetch(
+    `${env.NEXT_PUBLIC_API_URL}budgets/${id}/statistics?startDate=${startRange}&endDate=${endRange}`
+  );
 
   const { data: statistics, isLoading } = useQuery({
     queryKey: ["rangedStatistics", startRange, endRange, budget.id],
     queryFn: async () => {
-      return fetch(
-        `${env.NEXT_PUBLIC_API_URL}budgets/${id}/statistics?startDate=${startRange}&endDate=${endRange}`,
-        {
-          headers: {
-            accept: "*/*",
-            "Content-Type": "application/json",
-            Authorization: "Bearer " + session?.user.accessToken,
-          },
-        }
-      ).then((res) => res.json());
+      return fetch().catch((err) => console.error(err));
     },
-    enabled: !!session,
   });
   return (
     <StatisticsWrapperStyled>
--- a/apps/web/lib/hooks/useSuperfetch.ts
+++ b/apps/web/lib/hooks/useSuperfetch.ts
@@ -1,9 +1,13 @@
 import { useSession } from "next-auth/react";
 
+interface SuperOptions extends Omit<Omit<RequestInit, "headers">, "body"> {
+  body: object;
+}
+
 export default function useSuperfetch(
   url: string,
-  //disable ts suggestion for headers since we are setting them up manually
-  options?: Omit<RequestInit, "headers">
+  //disable ts suggestion for headers since we are setting them up manually and allow body to be object
+  options?: SuperOptions
 ) {
   const { data: session } = useSession();
 
--- a/apps/web/lib/hooks/useSuperfetch.ts
+++ b/apps/web/lib/hooks/useSuperfetch.ts
@@ -1,0 +1,33 @@
+import { useSession } from "next-auth/react";
+
+export default function useSuperfetch(
+  url: string,
+  //disable ts suggestion for headers since we are setting them up manually
+  options?: Omit<RequestInit, "headers">
+) {
+  const { data: session } = useSession();
+
+  const superfetch = () => {
+    const headers: HeadersInit = {
+      "Content-Type": "application/json",
+      accept: "*/*",
+      Authorization: "Bearer " + session?.user.accessToken,
+    };
+
+    return fetch(url, {
+      ...options,
+      body: JSON.stringify(options?.body),
+      headers,
+    })
+      .then((response) => {
+        if (response.ok) {
+          return response.json();
+        }
+      })
+      .catch((error) => {
+        throw new Error(`${error}`);
+      });
+  };
+
+  return superfetch;
+}
--- a/apps/web/lib/hooks/useSuperfetch.ts
+++ b/apps/web/lib/hooks/useSuperfetch.ts
@@ -23,10 +23,11 @@ export default function useSuperfetch(
       body: JSON.stringify(options?.body),
       headers,
     })
-      .then((response) => {
-        if (response.ok) {
-          return response.json();
+      .then((res) => {
+        if (res.ok) {
+          return res.json();
         }
+        //todo throw error based on status
       })
       .catch((error) => {
         throw new Error(`${error}`);
--- a/apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/BudgetBasicInformation.tsx
@@ -24,6 +24,12 @@ type BudgetBasicInfoProps = {
 };
 //TYPES end
 
+const TitleEditButton = styled.div`
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+  gap: 10px;
+`;
 export function BudgetBasicInformation({ budget }: BudgetBasicInfoProps) {
   const { t, dict } = useTranslate("BudgetsPage");
 
@@ -40,13 +46,6 @@ export function BudgetBasicInformation({ budget }: BudgetBasicInfoProps) {
       year: "2-digit",
     });
   }
-
-  const TitleEditButton = styled.div`
-    display: flex;
-    justify-content: flex-start;
-    align-items: center;
-    gap: 10px;
-  `;
 
   //DATA to display for information tiles
   const dataRangeInfo = (
--- a/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
@@ -8,6 +8,7 @@ import { ErrorMessage } from "ui";
 import { useSession } from "next-auth/react";
 import { Pagination } from "components";
 import { useTranslate } from "lib/hooks";
+import useSuperfetch from "lib/hooks/useSuperfetch";
 
 type APIResponse = {
   items: Item[];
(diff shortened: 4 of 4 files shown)


apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx
```diff
@@ -59,23 +60,15 @@ const BudgetStatistics = ({ budget }: Props) => {
...
-          headers: {
-            accept: "*/*",
-            "Content-Type": "application/json",
-            Authorization: "Bearer " + session?.user.accessToken,
-          },
-        }
-      ).then((res) => res.json());
+      return fetch().catch((err) => console.error(err));
     },
-    enabled: !!session,
```
Comment (mpiwonski) [inline apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx:78]: What happened to that line of code? What was the usage? Anyway, good job! :)

Comment (bdeluga) [reply apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx:78]: not needed now i guess 😕

apps/web/lib/hooks/useSuperfetch.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export default function useSuperfetch(
+  url: string,
+  //disable ts suggestion for headers since we are setting them up manually and allow body to be object
+  options?: SuperOptions
+) {
+  const { data: session } = useSession();
+
+  const superfetch = () => {
```
zuzannapaw, line 14: When I open your code in vscode, here is a suggestion
image

Since it has promises inside it, what do you think about it? 👀

bdeluga: Superfetch doesn't contain any async functions, it's fetch that does. So since it wouldn't change anything besides IDE happiness I would leave it as it is. But if you would like me to change it I'll be right happy to. 😄

apps/web/lib/hooks/useSuperfetch.ts
```diff
@@ -0,0 +1,38 @@
+import { useSession } from "next-auth/react";
+
+interface SuperOptions extends Omit<Omit<RequestInit, "headers">, "body"> {
+  body: object;
+}
```
Alessiass, line 5:
```suggestion
interface SuperOptions extends Omit<RequestInit, "headers"|"body"> {
  body: object;
}
```
Just suggestion :D

apps/web/lib/hooks/useSuperfetch.ts
Noshi96, line 14:
```suggestion
  const superfetch = (
    url: string,
    //disable ts suggestion for headers since we are setting them up manually and allow body to be object
    options?: SuperOptions
  ) => {
```

What do you think about making it more flexible (more super 😮) by passing these parameters inside superfetch?

apps/web/lib/hooks/useSuperfetch.ts
bdeluga: If result is the same I am happy with it 😄

apps/web/lib/hooks/useSuperfetch.ts
bdeluga: By passing them like this, we are marking `url` and `options` passed to hook as unused. And fetch won't work that way. 😞 

image

Noshi96: I'm talking about moving these attributes from the hook to the superfetch below 

Basically, it would look like this
```tsx
export default function useSuperfetch() {

  const superfetch = (
    url: string,
    //disable ts suggestion for headers since we are setting them up manually and allow body to be object
    options?: SuperOptions
  ) => {
```

and then the superfetch call changes slightly

```tsx
  const fetch = useSuperfetch();
  const URL = ``;
  const options = {};
  fetch(URL, options);
```
The main goal is preserved because you insert the token inside. The benefit I see in this approach is that we don't stick to the URL and option declarations in the functional component. This can be useful in a dynamically changing option body, for example in a tanstack queryFn

apps/web/lib/hooks/useSuperfetch.ts
```diff
@@ -0,0 +1,38 @@
...
+    return fetch(url, {
+      ...options,
+      body: JSON.stringify(options?.body),
+      headers,
+    })
+      .then(async (res) => {
+        if (res.ok) {
+          return { data: await res.json(), status: res.status };
+        }
+        return { data: null, status: res.status };
```
monika-klokosinska-intive, line 30: @bdeluga wouldn't it be better to throw an error here as well, so `superfetch` consumers could handle all problems in `onError` instead of slitting it between `onSuccess` and `onError`? @wlechowicz what do you think?

apps/web/lib/hooks/useSuperfetch.ts
bdeluga: You are right that would be more super to have and keep hook parameters empty.

apps/web/lib/hooks/useSuperfetch.ts
bdeluga: That would be better indeed, but JavaScript expects an error as a string. Unless you want me to pass the status as a string, I would have to create a CustomError to pass an object to it. Then, you would destructure the status anyway (due to the context of status throughout the whole app), which would be the same as casting the status string to a number.
### So...
It's up to the Team if we want to check for empty data to show some error. @intive/patronage-2023-js

apps/web/lib/hooks/useSuperfetch.ts
```diff
@@ -0,0 +1,38 @@
...
+      })
+      .catch((error) => {
+        throw new Error(`${error}`);
+      });
```
wyola, line 34: this catch seems redundant if you are rethrowing the error and doing nothing with it 🤔

apps/web/lib/hooks/useSuperfetch.ts
zuzannapaw: sure, I get it! thanks for response! 😄

apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx
```diff
@@ -105,10 +99,10 @@ const BudgetStatistics = ({ budget }: Props) => {
           />
           <CurrencyAmountStyled
             currencyOptions={currency}
-            amount={statistics?.periodValue}
+            amount={statistics?.data.periodValue}
```
Comment (wlechowicz) [inline apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx:102]: Why is this nested? It's effectively `data.data.stuff`

apps/web/lib/hooks/useSuperfetch.ts
wlechowicz: That's a bit awkward, since react-query already provides `{ data, error, isLoading, isError }` this should leverage that.
You shouldn't expect people to look at if !isError and then look again what status does `data` contain to make sure if it really is not an error.
If `!res.ok` then query should return `error`.
> The promise that is returned should either resolve the data or throw an error. For React Query to determine a query has errored, the query function **must** throw. Any error that is thrown in the query function will be persisted on the error state of the query.

https://tanstack.com/query/v3/docs/react/guides/query-functions#usage-with-fetch-and-other-clients-that-do-not-throw-by-default

I supposed for out use case, it would suffice to throw the status as a string, or indeed extend the Error class to have both the message, as well as the numerical error.

wlechowicz (changes requested):

apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx
Comment (wlechowicz) [reply apps/web/app/(navigation)/budgets/[id]/BudgetStatistics.tsx:102]: To avoid the awkward nesting of `data.data`, you can return it in a flat object like this:
```
if (res.ok) {
  const data = await res.json();
  return { ...data, httpStatus: res.status };
}
```

apps/web/lib/hooks/useSuperfetch.ts
monika-klokosinska-intive: I was thinking about throwing a new error object with status and message (since these two are potentially useful when handling errors), but even throwing status would be good!

apps/web/lib/hooks/useSuperfetch.ts
Noshi96, line 1: I tried to use supperfetch in useInfiniteQuery in SideNavigationBar.tsx. Without setting the "enabled" option and passing the session status inside, I couldn't prevent the 401 (Unauthorized) caused by the fetch when there is no session. This happens every time I refresh the page and open the sideNavBar. The solution I found is to return from the hook !!session as below and pass it to "enabled" option in queryFn. 
```tsx
return { superfetch, sessionStatus: !!session };
```

But this doesn't seem like a good solution to me. I don't know why this is happening. It might be related to the location of the SideNavigationBar.tsx call, app/(navigation)/layout. 🤷

bdeluga: I was trying to fix it but failed, simpler solution would be in case of InfiniteQuery to just useSuperFetch but to keep `enabled: !!session`, fetch would save us stringifying body, setting headers which is fair enough.

monika-klokosinska-intive (approved):

Alessiass (approved):
It works 🚀

wyola (approved):

Noshi96 (approved):

PR #97, opened 2023-05-17
feat: budget details - filtering by categories #P23-224

Noshi96:
## Filtering by categories ✍️ 

Ticket: https://tracker.intive.com/jira/browse/PATRO23-224

## Preview 👀 

https://patronage2023-js-web-git-feature-p23-224-beb6fb-patronage2023fe.vercel.app

## Overview - what was provided 🏗️ 🎯  

- Jotai atoms added
- Adjustments for side card with filter form
image
- The aside card now uses our checkboxes
- The aside card is now in the `sticky` position 
-  When the checkbox is changed, the list of transactions is updated (tanstack invalidate)
-  When the checkbox is changed, the pageIndex is reset
-  ~~New type for returned categories in HouseForm values `CategoryFilterType`~~


mpiwonski (approved):
Good job @Noshi96!

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -35,6 +37,11 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
   const { t, dict } = useTranslate("BudgetsPage");
   const setSorting = (column: string) => console.log(column);
   const { data: session } = useSession();
+  const categoryFilterState = useAtomValue(categoryFilterAtom);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
Comment (Noshi96) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:42]: This component renders every time the categoryFilterState changes, so it will work without the useEffect. This particular useEffect is responsible for resetting the pagination page when the categoryFilterState changes.

aniasgit (changes requested):
If you select a category and switch to a different budget, the new budget shows you data filtered by this category, although it is not marked that the category has been selected.

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -64,14 +71,23 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
...
+      categoryFilterState,
+    ],
     queryFn: async () => {
       return fetch(
         env.NEXT_PUBLIC_API_URL + "/budgets/" + budget.id + "/transactions",
         {
           body: JSON.stringify({
             pageSize: itemsPerPage,
             pageIndex: currentPage,
+            transactionType: null,
```
Comment (aniasgit) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:89]: I think transactionType can be removed from here. It's in my task, in a different version, adding it to yours will only cause unnecessary conflicts. According to swagger, not passing transactionType in the query will set it to null anyway on the BE side

apps/web/components/CategoryFilterForm/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  gap: 8px;
+`;
+
+const CategoryIconStyled = styled(CategoryIcon)`
+  width: 32px;
+  height: 32px;
+`;
```
wlechowicz, line 41: Put all this stuff in a separate file, please.

wlechowicz (approved):
LGTM, one refactor request

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (Noshi96) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx]: Done :D

apps/web/components/CategoryFilterForm/index.tsx
Noshi96: Done :D

apps/web/components/CategoryFilterForm/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const queryClient = useQueryClient();
+
+  return (
+    <Form
+      onSubmit={(values: CategoryFilterType) => {
+        const categoryFilterState: string[] = Object.entries(values)
+          .filter(([_, isChecked]) => isChecked)
+          .map(([categoryKeyName, _]) => categoryKeyName);
+
+        queryClient.invalidateQueries(["datatable"]);
```
monika-klokosinska-intive, line 30: @Noshi96 is it necessary to invalidate the query here? Won't the query re-run anyway because "categoryFilterState" changed?

aniasgit (approved):
Now it works good, great job 👍

PawelPazderski (commented):
What does the `manage` stand for? It's taking me back to:
image
Why is it here if the input selecting shows the filtering straight away?

Alessiass (approved):
Looks nice, works nice and utilizes Jotai. Kudos! 👻

Vectari (approved):
Sorry for complaining but I'm not sure about this yellow on yellow in "Refund". 
It stands out slightly from the others. Anyway works great!

zuzannapaw (commented):
The smaller card has become alive and finally dummy aside card content doesn't need to be imported! 🚀 

 A question from me: now, new category "Salary" has the same icon as one of the possible icons from icon picker for whole budget:   

image


image

I'm not sure it won't be confusing. What do you think about that?

apps/web/components/CategoryFilterForm/index.tsx
Noshi96: I agree with you that `queryClient.invalidateQueries` isn't needed in this situation. Already fixed. Thanks! :D

Noshi96:
> May I ask - What does the `manage` stand for? It's taking me back to: Why is it here if the selecting inputs shows the filtering straight away? Which looks and works nice by the way

@PawelPazderski 'manage' comes from the design. It looks like an unused element, some space filler. You are right that redirecting the user to the "/" page isn't the best solution. I changed it to a simple button with an empty call () =>{}. Thanks! :)

Noshi96:
> 



> Sorry for complaining but I'm not sure about this yellow on yellow in "Refund". It stands out slightly from the others. Anyway works great! 

Good point @Vectari! I have changed the contrast of the icon and its background

Noshi96:
> The smaller card has become alive and finally dummy aside card content doesn't need to be imported! 🚀
> 
> A question from me: now, new category "Salary" has the same icon as one of the possible icons from icon picker for whole budget:
> 
> image
> 
> image
> 
> I'm not sure if it won't be confusing. What do you think about that?

@zuzannapaw thanks for your comment! TBH, I feel that the icons available in the icon picker aren't that important. I believe that the icon, along with its name, displayed in the aside card, is sufficient for users to understand its representation and function. If you have any suggestions for alternative icons that could be used, I am open to considering them.

apps/web/components/CategoryFilter/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const categoryEntries = Object.entries(categoryMap);
+
+  const categoryFilterInit: CategoryFilterType = Object.keys(
+    categoryMap
+  ).reduce(
+    (acc, category): CategoryFilterType => ({
+      ...acc,
+      [category]: false,
+    }),
+    {} as CategoryFilterType
```
bdeluga, line 27: `Array.reduce` are allways complex another good solution would be 
```tsx
Object.fromEntries(Object.keys(categoryMap).map((key)=>[key,false])) as CategoryFilterType
```

bdeluga (approved):
LGTM 🚀 
One comment to maybe have chat about, not to change in particular.

apps/web/components/CategoryFilter/index.tsx
monika-klokosinska-intive: @bdeluga @Noshi96 another idea: just use array to store the selected categories. `[]` is the starting state (meaning no category was selected), once the user selected something, it is added to the array (e.g.  `['Subscriptions']` or whichever field is best used as key). 
If something is not in the array, it means it's not selected. There is a lot of code that transforms the object into an array and then back to object again. It will be simpler to just stick to an array all the time.

apps/web/lib/dictionary.tsx
```diff
@@ -871,6 +871,52 @@ const dictionary = {
...
+      fr: "Automobile",
+    },
+    grocery: {
+      en: "Grocery",
+      pl: "Zakupy spożywcze",
+      fr: "Épicerie",
+    },
+    salary: {
+      en: "Salary",
+      pl: "Pensja",
```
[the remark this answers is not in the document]
Noshi96: done :D

apps/web/components/CategoryFilter/index.tsx
Noshi96: Thanks, @bdeluga for showing an alternative to Array.reduce. @monika-klokosinska-intive I think your suggestion of using an array to store selected categories is very good indeed. I had complicated my solution. I have now applied your suggestions and I think it works better, thanks!

wyola (approved):

AzulloCyll (approved):

PR #99, opened 2023-05-18
feat: change language of interface #P23-29

AzulloCyll:
Task: https://tracker.intive.com/jira/browse/PATRO23-29

- added .svg flags to assets
- made language selector according to Jira story, which is flag image acting like select with dropdown menu

Used:
- RadixUI to build custom select with options
- Jotai to manage global state of current language
- localStorage interface to store selected language in web browser's memory

Addintional fixes:
- left menu have fixed width now

Todo:
- integrate language option with BE


apps/web/app/(navigation)/LanguageSelectorStyled.tsx
```diff
@@ -0,0 +1,52 @@
+import styled from "styled-components";
+import * as Select from "@radix-ui/react-select";
+
+export const SelectTriggerStyled = styled(Select.Trigger)`
+  background-color: unset;
+  cursor: pointer;
+  border: 0;
+  padding: 0;
+`;
```
wyola, line 9: add `line-height = 0` to fix this ⬇️ 

image

apps/web/app/(navigation)/LanguageSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { useHasScrollBar } from "lib/hooks/useHasScrollBar";
+
+export const LanguageSelector = () => {
+  const { hasScrollbar } = useHasScrollBar();
+  const setLanguage = useSetAtom(languageAtom);
+  const language = useAtomValue(languageAtom);
+
+  useEffect(() => {
+    const lang = localStorage.getItem("lang");
+    lang && setLanguage((localStorage.getItem("lang") as languages) || "en");
```
wyola, line 22: use the const :) 
```suggestion
    const lang = localStorage.getItem("lang");
    lang && setLanguage((lang as languages) || "en");
```
I would change the second line to use if statement. IMO it is more readable. Construction with `&&` is usually used if you want to assign the result to a variable. Also, if you already check if lang is true, then you don't need the `||`

```suggestion
    const lang = localStorage.getItem("lang");
    if (lang) {
        setLanguage(lang  as languages);
   }
```

apps/web/app/(navigation)/LanguageSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+    { lang: "en", flagSrc: "/flags/en.svg", languageName: "English" },
+    { lang: "fr", flagSrc: "/flags/fr.svg", languageName: "Français" },
+  ];
+
+  return (
+    <Select.Root
+      value={(language as languages) || "en"}
+      onValueChange={(lang) => {
+        changeLanguage(lang as languages);
+      }}>
```
wyola, line 41: you don't need casting here
```suggestion
    <Select.Root
      value={language || "en"}
      onValueChange={(lang) => {
        changeLanguage(lang);
      }}>
```

wyola (commented):
few improvements suggestions here and there 🤗

General question:
If we now keep language in `jotai` store then I don't think that we need `context` anymore 🤔 if I look for `LanguageContext` in the project, I see it is still used in a couple of places.

It means that we store language in two places and they are not in sync. I think that all usages of  `LanguageContext` should be replaced with `useAtomValue(languageAtom)`

Look what happens when I load the page in `fr` and then switch to `pl` - date picker uses `fr` language

image

mpiwonski:
image
I have noticed Log in/Log out buttons are not translated according to the language :)

apps/web/app/(navigation)/LanguageSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const changeLanguage = (lang: string) => {
+    setLanguage(lang as languages);
+    localStorage.setItem("lang", lang);
+  };
+
+  const items = [
+    { lang: "pl", flagSrc: "/flags/pl.svg", languageName: "Polski" },
+    { lang: "en", flagSrc: "/flags/en.svg", languageName: "English" },
+    { lang: "fr", flagSrc: "/flags/fr.svg", languageName: "Français" },
```
aniasgit, line 33: I would make the languages from the dictionary so that they are in the currently selected language. The flags are help enough for those who don't understand.

aniasgit: After discussion on Slack, I withdraw my comment, Daniel convinced me ;)

apps/web/app/(navigation)/LanguageSelector.tsx
aniasgit: How about: `setLanguage(lang ? lang as languages : "en")`

apps/web/app/(navigation)/LanguageSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const LanguageSelector = () => {
+  const { hasScrollbar } = useHasScrollBar();
+  const setLanguage = useSetAtom(languageAtom);
+  const language = useAtomValue(languageAtom);
+
+  useEffect(() => {
+    const lang = localStorage.getItem("lang");
+    lang && setLanguage((localStorage.getItem("lang") as languages) || "en");
+  }, [setLanguage, language]);
```
aniasgit, line 23: I think we only want useEffect once, so it should be empty array

zuzannapaw (commented):
I noticed that "currency" has not been translated 
image

 as well settings options: 
 
image


When we want to log in by this temporary button "Sign in",then translation does not work. It only works when we are using links to relocate to sign up/sign in. There is as well problem when I change url to relocate, then the page is not translated as well. 

image

image 

I am sure there is a solution for it! 💪 Anyway, application has become more alive with this feature, so well done 🎉

packages/ui/Flag/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+const ImgStyled = styled.img`
+  width: 2.3rem;
+  height: 1.5rem;
+  border-radius: 4px;
+  box-shadow: rgba(17, 17, 26, 0.05) 0px 1px 0px,
+    rgba(17, 17, 26, 0.1) 0px 0px 8px;
+`;
+
+export const Flag = ({ src }: FlagProps) => {
+  return <ImgStyled src={src} alt="" />;
```
monika-klokosinska-intive, line 16: @AzulloCyll alt shouldn't be empty here - you should pass it as another prop. Example "Flag of Poland", "Flag of UK" etc.

apps/web/app/(navigation)/LanguageSelector.tsx
monika-klokosinska-intive: @aniasgit `setLanguage` has to be included since it's used inside the hook, not sure why `language` itself is included.

Vectari:
"Add New Budget" button looks good in ENG and PL but FR:


Great job anyway!

Vectari (commented):
"Add New Budget" button looks good in ENG and PL but FR:


Great job anyway!

apps/web/app/(navigation)/LanguageSelectorStyled.tsx
AzulloCyll: All suggested changes are done! :)

apps/web/app/(navigation)/LanguageSelector.tsx
AzulloCyll: done

apps/web/app/(navigation)/LanguageSelector.tsx
AzulloCyll: Changed

apps/web/app/(navigation)/LanguageSelector.tsx
AzulloCyll: Changed

packages/ui/Flag/index.tsx
AzulloCyll: All done

wyola (approved):
LGTM! 🚀

zuzannapaw (approved):
Well done with changes! 🚀  One small thing for me, because buttongroup's text is long in french version 🐸 , it would be nice to add some padding to it, or set words not under, but next to each other: 

image

Also text in aside card is not being translated, but maybe a good thing would be to wait for this card's content when sorting by categories functionality will be merged. Other than that, it looks good to me ☺️

bdeluga (approved):
Everything works superb, Approve 🚀 
I would do something about that button group @zuzannapaw mentioned.

apps/web/app/store.tsx
bdeluga, line 1: I don't feel like using enums in TS, they are weird. We could simplify this and use type instead:

```ts
export type Language = "en" | "pl" | "fr";
```
This achieves same point but it's simpler 🚀

apps/web/app/layout.tsx
```diff
@@ -22,6 +24,13 @@ const inter = Inter({
 const queryClient = new QueryClient();
 
 export default function RootLayout({ children }: LayoutProps) {
+  const setLanguage = useSetAtom(languageAtom);
+
+  useEffect(() => {
+    const lang = localStorage.getItem("lang") as languages;
+    setLanguage(lang ? (lang as languages) : ("en" as languages));
+  }, [setLanguage]);
+
```
bdeluga, line 33: We were discussing this article that @wlechowicz sent about [paralel routes next.js](https://nextjs.org/docs/app/building-your-application/routing/parallel-routes). There is a logic in Layout which we don't like, simply because it's not a layout job to share state or do other things than _layout_ related.
We can move this logic to language selector itself.

apps/web/app/(navigation)/LanguageSelector.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export const LanguageSelector = () => {
+  const { hasScrollbar } = useHasScrollBar();
+  const setLanguage = useSetAtom(languageAtom);
+  const language = useAtomValue(languageAtom);
+
+  const changeLanguage = (lang: string) => {
+    setLanguage(lang as languages);
+    localStorage.setItem("lang", lang);
+  };
```
bdeluga, line 21: I explained why this suggestion occured on `apps/web/app/layout.tsx` file


```suggestion
  const changeLanguage = (lang: string) => {
    setLanguage(lang as languages);
    localStorage.setItem("lang", lang);
    
      useEffect(() => {
    //localStorage.getItem("lang") works as well
    setLanguage(localStorage.lang as Language);
  }, [setLanguage]);
    
  };
```

bdeluga (changes requested):
Still works superb but there are a few requests 🔨

apps/web/app/layout.tsx
AzulloCyll: That logic was in languageSelector but...

i've moved it there beacase of that our languageSelector exists only on (navigation) layout, so our (regflow) remains untranslated.

bdeluga: Right then... @wlechowicz what should we do then ?

AzulloCyll: done

apps/web/app/(navigation)/LanguageSelectorStyled.tsx
```diff
@@ -0,0 +1,53 @@
...
+  border: 0;
+  padding: 0;
+  line-height: 0;
+`;
+
+export const SelectContentStyled = styled(Select.Content)`
+  overflow: hidden;
+  color: ${({ theme }) => theme.datePicker.neutral10};
+  background-color: ${({ theme }) => theme.card.background};
+  z-index: 40000;
```
monika-klokosinska-intive, line 16: @AzulloCyll does this really need to be this big?!

Vectari (approved):
🚀

apps/web/app/(navigation)/LanguageSelectorStyled.tsx
AzulloCyll: Changed z-index value to 10

bdeluga (approved):
LGTM 🚀

apps/storybook/stories/Flag.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+export default {
+  title: "Flag",
+  component: Flag,
+} as ComponentMeta<typeof Flag>;
+
+const Template: ComponentStory<typeof Flag> = ({ ...args }) => (
+  <Flag {...args} />
+);
+
+// import PL from "../../web/public/flags/pl.svg";
```
monika-klokosinska-intive, line 14: @AzulloCyll please remove this comment if it isn't necessary anymore!

PawelPazderski (approved):
Works like charm, one comment reffering to conversation about radix, other than that 🚀

Noshi96 (approved):

PR #100, opened 2023-05-18
feat: budget details - filtering by type #P23-223 

aniasgit:
task: [#P23-223](https://tracker.intive.com/jira/browse/PATRO23-223)


packages/ui/ButtonGroup/index.tsx
```diff
@@ -137,3 +122,24 @@ const ButtonGroupStyled = styled.div<GroupProps>`
...
+  return (
+    <ButtonGroupStyled options={options} secondary={secondary}>
+      {options.map(({ component, onSelect, defaultChecked, id }, index) => {
+        return (
+          <Fragment key={id}>
+            <input
+              type={"radio"}
+              id={`button-${index}`}
+              onClick={onSelect}
+              name={"button-group"}
```
wyola, line 136: you don't need brackets when passing strings
```suggestion
              type="radio"
              id={`button-${index}`}
              onClick={onSelect}
              name="button-group"
```

packages/ui/ButtonGroup/index.tsx
```diff
@@ -137,3 +122,24 @@ const ButtonGroupStyled = styled.div<GroupProps>`
...
+
+export const ButtonGroup = ({ options, secondary }: GroupProps) => {
+  return (
+    <ButtonGroupStyled options={options} secondary={secondary}>
+      {options.map(({ component, onSelect, defaultChecked, id }, index) => {
+        return (
+          <Fragment key={id}>
+            <input
+              type={"radio"}
+              id={`button-${index}`}
```
wyola, line 134: `id` should not be based on `index`. If there were two button groups on one page, then in both of them there would be a button with for example `id="button-0"`.
```suggestion
              id={`button-${id}`}
```

apps/web/lib/dictionary.tsx
```diff
@@ -710,6 +710,11 @@ const dictionary = {
         fr: "Solde total",
       },
     },
+    buttonGroupLabels: {
+      all: { en: "View all", pl: "Wszystkie", fr: "Voir tout" },
+      income: { en: "Income", pl: "Wpływy", fr: "Les recettes" },
```
wyola, line 715: I know design says "Income" but since `pl` and `fr` are in plural form here, and `Expenses` below are also in plural, I would also change it here.
```suggestion
      income: { en: "Incomes", pl: "Wpływy", fr: "Les recettes" },
```

wyola (commented):
Few suggestion, other than that looks good!

I would add `defaultChecked` to `View all`, so when user enters the page, then this option is selected. Now nothing is selected, yet transactions are rendered and it looks a bit strange IMO

image

packages/ui/ButtonGroup/index.tsx
aniasgit: it's not my code, I just moved the styled components to the top, but I can improve it of course :)

packages/ui/ButtonGroup/index.tsx
aniasgit: same as: https://github.com/intive/patronage2023-js/pull/100#discussion_r1197623477

apps/web/lib/dictionary.tsx
aniasgit: I agree :)

packages/ui/ButtonGroup/index.tsx
wyola: I know,  just caught my eye :) would be good to improve it anyway if you have a moment 🤗

packages/ui/ButtonGroup/index.tsx
aniasgit: done

packages/ui/ButtonGroup/index.tsx
aniasgit: done

apps/web/lib/dictionary.tsx
aniasgit: done

aniasgit:
> image There is probably some minor tweak about width. Other than that LGTM 🚀

@bdeluga check it now

bdeluga (approved):
Now it's perfect !

mpiwonski (approved):
Good job :)

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -91,6 +106,33 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
...
+      id: "income",
+    },
+    {
+      component: t(dict.buttonGroupLabels.expenses),
+      onSelect: () => {
+        setTransactionTypeFilterState("Expense");
+      },
+      id: "expenses",
+    },
+  ];
```
Comment (wlechowicz) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:132]: This is fogging up the controller component, please put this in a separate component that keeps this config list, uses ButtonGroup and accepts a setTransactionType through props, so the controller can just import that, put it in the JSX and pass the setter to it, along the currently selected value, via props.

wlechowicz (changes requested):

Alessiass (approved):
Looks good and works good. At the beginning I was thinking that useEffect for setting page was unnecessary because it is built in Pagination but I was wrong.
Good job! 👍

monika-klokosinska-intive (approved):

PR #103, opened 2023-05-20
feat: dictionary validator #P23-344

bdeluga:
# Introducing dictionary validator ✨ 
Ever found yourself not translating texts of components and sticking to english, or maybe you didn't bother translating to that 🐸 French language. Well too bad, because from now on, app won't build when there are missing translations in our dictionary.

## Usage ✏️ 
If you want to run it manually:
```bash
npm run dict-check
```
Reminder: **Check will automatically run before every build.** 🚩

## Validation message 🔔 
There are two things that could happen
- you forgot translation

image
- or maybe more than one

image

- you are up to date with translations (nice 👍)

image

## Discussion 📖 
I've installed `cli-color` which is purely for keeping it "pretty".  @intive/patronage-2023-js Give your feedback about coloring
\- do we need it ? Because it's a `51.4k (gzipped 15.9k)` just for text to be colored. We can get rid of it.
Also @wlechowicz I assumed thats how prebuild check would work, feel free to change what you need to make it work.

## Scalability 📈 
Need to add spanish language ? Simply add it to hasMissingProperties function.

```tsx
//dict-check.ts

//["en", "pl", "fr"] default setting
const hasMissingProperties = (obj: DictionaryValue): string[] =>
  ["en", "pl", "fr","es"].filter((key) => !(key in obj));


```


PawelPazderski (approved):
Wow, works great, congratulations 🚀

AzulloCyll (commented):
It searching for all keys to exesist in our dict object. 
Shouldn't check for empty string as a value also?

bdeluga:
> It searching for all keys to exesist in our dict object. 
> 
> Shouldn't check for empty string as a value also?
> 
> 
> 
> 

Empty strings won't break our app. I was aiming to remove possibility of undefined to happen.

zuzannapaw (approved):
I'm truly impressed by this! 👀 It works great for a single and for multiple lacks of language. Nice thought with adding some colors to messages. Well done ✨

PR #105, opened 2023-05-21
feat: budget details searching by name #P23-222

Vectari:
# Hello!
[Task P23-222 - Budget details - searching by name](https://tracker.intive.com/jira/browse/PATRO23-222)

### Have a nice day!


bdeluga (approved):
Nice touch with debounce. Approved🚀

zuzannapaw (approved):
Well done! Works great! ✨  I couldn't reach it from pr, but only one, small thing that I would do, is keep these ui's imports under each other for code neatness. 

image

Congrats! 🚀

aniasgit (changes requested):
During the search, we should change the current page in the pagination to 1. If I enter something on the 3rd page that returns results on only one page, it shows me data not found, I see the results only when I click on 1 in the pagination.

Alessiass (approved):
LGTM 🚀

Vectari:
@aniasgit You are right. Now page back to 1 after search and everything work fine. Thank you for review!

Vectari:
@zuzannapaw @monika-klokosinska-intive I hope is better now. Thank you for review 🙂

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -129,6 +134,13 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
     <>
       <FilterSearchWrapper>
         <TransactionTypeFilter onSelect={(type) => setTransactionType(type)} />
+        <SearchInput
+          placeholder={`${t(dict.searchInputTransactionPlaceholder)}`}
+          onChange={(e) => {
+            setSearchTransactionByName(e.currentTarget.value);
+            setCurrentPage(1);
```
Comment (aniasgit) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:141]: It works fine now, but there are two useEffects (on line 51 and 122) that do the same thing, you can combine them, i.e. make one with a dependency array that will have values from both and add to that array, your debouncedSearch and remove this setCurrentPage from here.

aniasgit (approved):
I'm still asking you for a little change.

PawelPazderski (approved):
LGTM, works well 🚀 
Just wondering about BE side here, why the search works if matches any part of string here (like "sh" filters "Cash") while in Users List it works when matching only from the begining of the string (so "sh" will show no data in this example, only if we type c..a... and so on would)

monika-klokosinska-intive (changes requested):
Please take a look at @aniasgit's suggestion.

AzulloCyll (approved):
🚀

wlechowicz (approved):

wlechowicz (approved):

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (Vectari) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx]: Good advice. I changed it 👍

Vectari:
@aniasgit @monika-klokosinska-intive I added dependency array and hope is alright now 😄

Noshi96 (approved):
LGTM 🚀

monika-klokosinska-intive (approved):

PR #106, opened 2023-05-21
refactor: resize avatars #P23-345

wyola:
## Decrease the size of avatars' files

Biggest dimensions of the avatar in the design are 120 x 120 px. I resized the avatars to be 240 x 240 px so they display well on retina screens.

image

I removed not used avatars from `public`. I changed some components/stories to use avatars from `public/avatars`, so we don't have to keep additional avatar just for those. Let me know if that's okay.


bdeluga (approved):
Nice job ! 🚀

zuzannapaw (approved):
Good job! ⭐ Nice change with src paths in stories and components, it gives us more order.

AzulloCyll (approved):
LGTM! 🚀

monika-klokosinska-intive (approved):

aniasgit (approved):

PR #109, opened 2023-05-22
feat: users list in administrator menu #P23-25

PawelPazderski:
## Users list displayed **_when logged as administrator_** - available in side menu - Users

Ticket: https://tracker.intive.com/jira/browse/PATRO23-25

The list displays users data:

- avatar - if the avatar is not available the default avatar is shown 

( at the moment it's checked if the avatar string provided by BE either matches a position in the public/avatars or if it's an url address )
- last name - sorted by default
- first name
- email
- date the user was created

In the mobile view only name & email stays shown.

You can sort all those columns using the buttons in table header or search using the search input which works around last name, first name and email columns.


bdeluga (commented):
If you could try convincing BE to fix those pesky avatars. 🚀

apps/web/app/(navigation)/users/UsersList.tsx
```diff
@@ -0,0 +1,150 @@
...
+                  if (!props.rowData.avatar) return <Avatar src="/unsetAvatar.svg" />
+
+                  const text = props.rowData.avatar;
+                  //check if avatar string is one from public folder
+                  const schemaPath = z.string().startsWith("/avatars/").endsWith(".svg");
+                  //check if avatar url is valid url
+                  const schemaUrl = z.string().url();
+                  const isPath = schemaPath.safeParse(text);
+                  const isUrl = schemaUrl.safeParse(text);
+                  return <Avatar src={isPath.success || isUrl.success ? props.rowData.avatar : "/unsetAvatar.svg"} />;
```
bdeluga, line 108: This shouldn't be done this way. Let's suppose we have avatars that have `.png` extension, then we get broken avatars, because we only check for and `.svg` extension. `/avatars/` could also be moved/renamed. I know it's a workaround for not getting right avatars from backend, but maybe you could ask them to change users names to `""` (empty string) where there are still users with weird images `src`.

Then, because `""` is `falsy` it would allow us to do 
```tsx
<Avatar src={props.rowData.avatar || "/unsetAvatar.svg"} />
```

apps/web/app/(navigation)/users/UsersList.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { Column } from "ka-table/models";
+
+import { Icon, Avatar } from "ui";
+import { z } from 'zod';
+
+import {
+  UsersListStyled,
+  EmailStyled
+} from "./UsersList.styled";
+import { TransactionsTableSuspense } from "./../budgets/[id]/TransactionsTableSuspense";
```
zuzannapaw, line 15: To avoid confusion with naming, you could change this import to `import { TransactionsTableSuspense as UsersTableSuspense } from "./../budgets/[id]/TransactionsTableSuspense";` .Since this suspense is not being imported for the transactions table, with this change it will be easier to look through code ☺️

apps/web/app/(navigation)/users/UsersList.styled.tsx
```diff
@@ -0,0 +1,178 @@
...
+  .ka-group-cell {
+    padding: 8px 8px 8px 0;
+  }
+
+  .ka-group-cell-content {
+    font-size: 12px;
+    font-weight: 400;
+    line-height: 16px;
+    color: ${({ theme }) => theme.transactionsTable.date};
+  }
```
aniasgit, line 119: I don't think you need it

apps/web/app/(navigation)/users/UsersList.tsx
```diff
@@ -0,0 +1,148 @@
+"use client";
+
+import { useContext } from "react";
+import { ThemeContext } from "styled-components";
+import { useTranslate } from "lib/hooks";
+import { Table } from "ka-table";
+import { DataType } from "ka-table/enums";
+import { Column } from "ka-table/models";
+import { Icon, Avatar } from "ui";
```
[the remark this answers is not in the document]
PawelPazderski: I do here, for sorting buttons in the header 
image

apps/web/app/(navigation)/users/UsersList.styled.tsx
PawelPazderski: yeah, there are some unneeded leftovers from the TransactionTable styling, I'll check on that

apps/web/app/(navigation)/users/UsersList.tsx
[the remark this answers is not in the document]
aniasgit: Ok 👍

zuzannapaw (commented):
Great job! 🚀  

Few things from me, I see the problem with line under table headers :
image

You can fix this by minimizing "left" for -27px on your ` ::after` line element 
image

`.TiBQr .ka-thead::after {
    content: " ";
    position: absolute;
    background: #E1E1E1;
    height: 1px;
    bottom: 0;
    left: -27px;
    right: -54px;}`

But the bigger problem occurs on mobile size. It looks like "Name" header does not have enough space, also whole card is not responsive here.

image

Although nice adjustement that on mobile size only 3 columns are visible! ✨

apps/web/app/(navigation)/users/UsersList.tsx
```diff
@@ -0,0 +1,148 @@
...
+                  //check if avatar string is valid url
+                  const schemaUrl = z.string().url();
+                  const isPath = schemaPath.safeParse(text);
+                  const isUrl = schemaUrl.safeParse(text);
+                  return <Avatar src={isPath.success || isUrl.success ? rowData.avatar : "/unsetAvatar.svg"} />;
+                case "email":
+                  return <EmailStyled>{rowData.email}</EmailStyled>;
+                case "createdTimestamp":
+                  const date = new Date(rowData.createdTimestamp);
+                  const dateString = `${date.getFullYear()}-${((date.getMonth() + 1) < 10 ? '0' : '') + (date.getMonth() + 1)}-${(date.getDate() < 10 ? '0' : '') + date.getDate()}`;
```
aniasgit, line 111: I'd rather use this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

PawelPazderski: The idea here is to show the data as in design which I assume is same for all language versions (year - month - day) - couldn't see that in localeDateString

apps/web/lib/dictionary.ts
```diff
@@ -576,7 +576,7 @@ const dictionary = {
         en: "Oops, something went wrong.",
         pl: "Ups, coś poszło nie tak.",
         fr: "Oups, quelque chose s'est mal passé.",
```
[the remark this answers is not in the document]
PawelPazderski: ok sure

apps/web/app/(navigation)/users/UsersTableController.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { useState, useEffect } from "react";
+import { env } from "env.mjs";
+import { useQuery } from "@tanstack/react-query";
+import { ErrorMessage } from "ui";
+import { useSession } from "next-auth/react";
+import { Pagination } from "components";
+import { useTranslate } from "lib/hooks";
+import { SearchInput } from "ui/Input/SearchInput";
+import { InputWrapper } from "./UsersList.styled";
+import { useDebounce } from "lib/hooks/useDebounce";
```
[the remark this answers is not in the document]
PawelPazderski: ok, sounds like a good idea

apps/web/app/(navigation)/users/UsersList.tsx
PawelPazderski: ok, done

AzulloCyll:
@zuzannapaw 

I've adjusted left menu on my branch with languagePicker

apps/web/app/(navigation)/users/UsersList.tsx
wlechowicz: npm date-fns format

AzulloCyll (commented):
Please resolve conflicts and accept languageContext file deletion. :)

I've checked how it's working with languageSelector and it's good partially.
Header name of the page is changing language.
But I don't know about column names - they not changing instantly after changing language, but only when we hit F5.

PS. It occurs on budget's transaction table too!

monika-klokosinska-intive:
@PawelPazderski @AzulloCyll An idea on how to fix the table columns not refreshing
Instead of running the t function inside the table, let's run it's outside of it and use the result inside the table.
```
const firstNameTranslated = t(usersTable.lastName);
```
```
  {
      key: "lastName",
      title: firstNameTranslated,
      isSortable: true,
      dataType: DataType.String,
      style: { width: "30%" },
    },
```

I am thinking this might be a ka-table problem - it creates columns once and ignores changes afterwards... let's see if moving the translating outside of the table will force the table to change it.

apps/web/app/(navigation)/users/UsersList.tsx
PawelPazderski: OK, I'll use that, thanks

monika-klokosinska-intive:
@PawelPazderski oh man, no more ideas so far :(.

apps/web/app/(navigation)/users/UsersTableController.tsx
```diff
@@ -0,0 +1,138 @@
...
+        isLoading={isLoading}
+      />
+      <Pagination
+        pageIndex={currentPage - 1}
+        numberOfPages={totalPages}
+        pageSizeOptions={[5, 10, 25]}
+        currentPageSize={itemsPerPage}
+        onChangePageSize={(val) => {
+          setItemsPerPage(val);
+          setCurrentPage(1);
```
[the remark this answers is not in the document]
PawelPazderski: I don't quite follow up here, I use the pagination component same way it is used in TransactionTable, can you explain please

PawelPazderski:
> @PawelPazderski oh man, no more ideas so far :(.

I'm on it like 5th hour now and nothing so far, this russian technology drives me mad slowly

apps/web/app/(navigation)/users/UsersTableController.tsx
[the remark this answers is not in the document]
PawelPazderski: Ok never mind got it

monika-klokosinska-intive:
@PawelPazderski it is okay to merge it without this issue fixed, since it's app-wide and not related to only your task. Looks like ka-table was too good to be true.

PawelPazderski:
> @PawelPazderski it is okay to merge it without this issue fixed, since it's app-wide and not related to only your task. Looks like ka-table was too good to be true.

We made it actually, Ola came in all dressed up white and saved the day :)

Vectari (approved):
Works great, good job! 🥇 Only one thing, maybe you should use superFetch.

apps/web/app/(navigation)/users/UsersList.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+  return (
+    <UsersListStyled>
+      <Table
+        key={language}
+        columns={columns}
+        rowKeyField={"id"}
+        data={users}
+        noData={{
+          text: "No Data Found",
```
aniasgit, line 103: It should come from dictionary, please change it also in transactionsTable.

apps/web/app/(navigation)/users/UsersList.tsx
```diff
@@ -0,0 +1,190 @@
...
+          cell: {
+            content: ({ column, rowData }) => {
+              switch (column.key) {
+                case "avatar":
+                  //set default avatar if missing data
+                  if (!rowData.avatar) return <Avatar src="/unsetAvatar.svg" />;
+
+                  const text = rowData.avatar;
+                  //check if avatar string is a path coming from our avatars folder
+                  const schemaPath = z.string().startsWith("/avatars/");
```
aniasgit, line 115: what's happen when path will be for example "/avatars/10.svg" and we don't have nothing like this in our avatars folder? I appreciate you trying to check if we have the correct avatar, but your solution is not perfect.

apps/web/app/(navigation)/users/UsersTableController.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const {
+    data: users,
+    isError,
+    isLoading,
+    refetch,
+    error,
+  } = useQuery({
+    queryKey: ["user", itemsPerPage, currentPage, sortParams, debouncedSearch],
+    queryFn: async () => {
+      return fetch(`${env.NEXT_PUBLIC_API_URL}user/list`, {
```
aniasgit, line 75: since we have a superfetch, it should be used here ;)

apps/web/app/(navigation)/users/UsersList.tsx
PawelPazderski: We won't have "/avatars/10.svg" because that information comes from us at the sign up - this wolud mean our up allows to choose avatar we don't have, my solution here is for protection against unwanted strings which we can't serve and are present on list (which is good for now because allows us to show the default avatar)

PawelPazderski:
> @PawelPazderski please address Ania's two comments - ("no data") coming from dictionary and super-fetch and we are good to go!

I'm on it

apps/web/app/(navigation)/users/UsersTableController.tsx
PawelPazderski: done, thanks

monika-klokosinska-intive (approved):

zuzannapaw (approved):
Good job! 🚀

Noshi96 (approved):

PR #110, opened 2023-05-22
feat: choices memoization #P23-336

Alessiass:
## Overview 📖 

Import hook:

`import { useLocalStorage } from "lib/hooks";`

Use it as useState!

`const [getPageSizeValue, setPageSizeValue] = useLocalStorage("pageSize", "6");`

Keep in mind to parse stored values from string.

### DIFF
--- a/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
@@ -85,6 +85,7 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
             pageSize: itemsPerPage,
             pageIndex: currentPage,
             transactionType: transactionType,
+            search: "",
           }),
           headers: {
             Authorization: "Bearer " + session!.user.accessToken,
--- a/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
@@ -33,7 +33,6 @@ type ID = {
 const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
   const [getPageSizeValue, setPageSizeValue] = useLocalStorage("pageSize", "6");
   const [currentPage, setCurrentPage] = useState<number>(1);
-  const [itemsPerPage, setItemsPerPage] = useState<number>(5);
   const [totalPages, setTotalPages] = useState<number>(1);
   const [transactionType, setTransactionType] = useState<
     "Income" | "Expense" | null
--- a/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
@@ -31,7 +31,7 @@ type ID = {
 };
 
 const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
-  const [getPageSizeValue, setPageSizeValue] = useLocalStorage("pageSize", "6");
+  const [getPageSizeValue, setPageSizeValue] = useLocalStorage("pageSize", "5");
   const [currentPage, setCurrentPage] = useState<number>(1);
   const [totalPages, setTotalPages] = useState<number>(1);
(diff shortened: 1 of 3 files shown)


apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -49,7 +49,7 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
   }, [categoryFilterState]);
 
   const fixFetchedData = (res: APIResponse) => {
-    setTotalPages(Math.ceil(res.totalCount / itemsPerPage));
+    setTotalPages(Math.ceil(res.totalCount / parseFloat(getPageSizeValue)));
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:52]: @Alessiass any chance of using `parseFloat` once, saving the result and re-using it, instead of repeating the parsing each time it's needed in this file?

monika-klokosinska-intive (approved):

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (Alessiass) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx]: Done :)

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -43,13 +43,14 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
   const setSorting = (column: string) => console.log(column);
   const { data: session } = useSession();
   const categoryFilterState = useAtomValue(categoryFilterAtom);
+  const pageSize = parseFloat(getPageSizeValue);
```
Comment (wyola) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:46]: this should be changed to `parseInt`. it is kept in local storage, so it can be changed by user. right now, I can change it to 4.5 and it is sent like this to backend. Transaction table crashes then.

```suggestion
  const pageSize = parseInt(getPageSizeValue);
```


image

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -33,8 +33,8 @@ type ID = {
 };
 
 const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
+  const [getPageSizeValue, setPageSizeValue] = useLocalStorage("pageSize", "5");
```
Comment (wyola) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:36]: I think that name `pageSize` is too generic and it can clash with other tables in the future, for example users table. Maybe it would be better to name it `transactionsTablePageSize`.

wyola (commented):
Two remarks from me, other than that I think it looks good!

Vectari (approved):

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (monika-klokosinska-intive) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx]: @Alessiass great job! Please also address @wyola excellent point of using `parseInt` instead ;).

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (Alessiass) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx]: Both suggestions are valid and well thought, we've faced this problem before with queries. Implemented 🚀

bdeluga (approved):
LGTM :rocket:

apps/web/lib/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,21 @@
...
+    let currentValue;
+    try {
+      currentValue = JSON.parse(
+        localStorage.getItem(key) || String(defaultValue)
+      );
+    } catch (error) {
+      currentValue = defaultValue;
+    }
+    return currentValue;
+  });
```
wlechowicz, line 14:
```suggestion
    try {
      return JSON.parse(
        localStorage.getItem(key) || String(defaultValue)
      );
    } catch (error) {
      return defaultValue;
    }
  });
```

wlechowicz (changes requested):

wlechowicz (approved):

PR #114, opened 2023-05-23
feat: people in budget #P23-219

wyola:
## People in budget 🧑‍🤝‍🧑 
task: https://tracker.intive.com/jira/browse/PATRO23-219

Now users can see who they share the budget with!

- logged in user is not displayed in avatars
- avatars are stacked up to 4 users
- above 4 users, the last spot displays a grey circle with the number of remaining users
- I added a new atomic component - a tooltip (made with radix). I was going to use an HTML `title` attribute but it did not look good, especially when hovering over the gray circle avatar aggregator - users were displayed in one line and there is no way of styling it. Client accepted how the radix tooltip looks.
- I worked out some styles for mobile 🫠 there is a new breakpoint, which is specific to this screen, but it is a lot in styles, so I just added it in media-queries for now. Let's not focus on that, since like we discussed - the top part's layout has to be redone anyway.

### Test
I shared in the PR's thread on Slack credentials where you can log in and see various cases
https://intive-apu8934.slack.com/archives/C04R6LM4JSV/p1684873404585919

### Issue
What should we do if there is a broken avatar incoming from backend? I don't think there is a way of this happening, but should there be some fix for it anyway? Same issues is with users' table or with creator in transactions' table

image


apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const StyledCounter = styled.div`
+  border: 2px solid ${({ theme }) => theme.avatar.outline};
+  background-color: ${({ theme }) => theme.avatar.aggregator};
+  color: ${({ theme }) => theme.avatar.outline};
+  border: 2px solid ${({ theme }) => theme.avatar.outline};
+  border-radius: 50%;
+  width: 1em;
+  height: 1em;
+  display: flex;
```
Comment (Noshi96) [inline apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:35]: So basically 1em here is 48px in the desktop view because it was inherited from StyledWrapper, which is confusing imo. And also, when we use "em" in `StyledWrapper` for `line-height`, shouldn't we also use it for `font-size`? What do you think about setting `font-size` in `StyledWrapper` to:
```tsx
  ${device.tablet} {
    font-size: 1em;
    height: 48px;
  }
```
and then:
 ```css
  width: 3em;
  height: 3em;
```
In this case, we must also directly set the `font-size` for the avatars:
```tsx
const StyledAvatar = styled(Avatar)`
  font-size: 3em;
`;
```

Noshi96 (commented):
Great job! The user with prepared budgets was very helpful during testing, as was the PR description. I like the new tooltip. It looks and works great! Here below are some minor changes I would make.

packages/ui/Avatar/index.tsx
```diff
@@ -21,19 +22,29 @@ const Image = styled.img<AvatarProps>`
...
-      src={src}
-      alt={alt ? alt : `${username}'s avatar`}
-      className={className}
-      outlined={outlined}
-    />
-  );
-};
+export const Avatar = forwardRef<HTMLImageElement, AvatarProps>(
+  (
+    {
```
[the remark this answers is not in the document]
wyola: @Noshi96 you are right, I don't know why my IDE did not shout at me, I only see it now after running `npm run lint` 😢  I will fix it, thank you!

image

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:35]: @Noshi96 not really, because the way the Avatar component was designed, was to set the `font-size` and `line-height` on it's parent, and then Avatar has `1em` and inherits the `font-size`. The way it is right now you set the `font-size` only in `StyledWrapper` and don't worry about the rest. I followed the same approach for styling the counter.

The way you propose will cause that if you want to change something you have to do it separately for Avatar and the counter, instead of changing it in only one place.

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
```diff
@@ -0,0 +1,94 @@
...
+  if (peopleWithoutLoggedUser.length > 4) {
+    shortUserList = peopleWithoutLoggedUser.slice(0, 3);
+    remainingUsers = peopleWithoutLoggedUser.slice(3);
+  }
+
+  const remainingUserNames = remainingUsers.map((user) => (
+    <StyledUser key={user.id}>
+      {user.firstName} {user.lastName}
+    </StyledUser>
+  ));
```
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:67]: @Noshi96 if you feel strongly about it I can change it but I think it is a kind of "potayto, potahto" situation 🍟

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return (
+    <StyledWrapper>
+      {shortUserList.map((user) => (
+        <Tooltip
```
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:72]: like here https://github.com/intive/patronage2023-js/pull/114#discussion_r1203782079

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
```diff
@@ -0,0 +1,94 @@
...
+  padding: 3px;
+`;
+
+const PeopleInBudget = ({ budget }: PeopleInBudgetProps) => {
+  const { data: session } = useSession();
+
+  const loggedUser = session?.user.id;
+  const peopleWithoutLoggedUser = budget.budgetUsers.filter(
+    (user) => user.id !== loggedUser
+  );
```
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:54]: @aniasgit okey-dokey, I will do it for you

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  let shortUserList = peopleWithoutLoggedUser;
```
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:56]: @zuzannapaw good idea, sounds better! I will change it though to simply `visibleUsers`, then we have `visibleUsers` and `remainingUsers` so it is more consistent 🤗

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  ${device.tablet} {
+    font-size: 48px;
+  }
+`;
+
+const StyledCounter = styled.div`
+  border: 2px solid ${({ theme }) => theme.avatar.outline};
+  background-color: ${({ theme }) => theme.avatar.aggregator};
```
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:29]: @zuzannapaw I know, but it comes from the design so 🤷🏻‍♀️

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
```diff
@@ -0,0 +1,94 @@
+import styled from "styled-components";
+import { BudgetFixed, BudgetUser } from "lib/types";
+import { useSession } from "next-auth/react";
+import { Avatar, Tooltip } from "ui";
+import { device } from "lib/media-queries";
```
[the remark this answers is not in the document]
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:5]: `useSession` was moved elsewhere to match this request https://github.com/intive/patronage2023-js/pull/114#discussion_r1203741097 and I moved the `ui` import to the bottom

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
Comment (Noshi96) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:34]: My solution also works for the Avatar and the counter, but has the opposite logic. After reconsidering, I see more advantages to your solution, so let's keep them. I'm still thinking about changing, px to em for font sizes.

Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:34]: @Noshi96 I will be redoing layout of the top anyway, so I will revisit this topic then :) because most likely some things will change here anyway

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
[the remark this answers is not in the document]
Comment (AzulloCyll) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx]: that runtime errro is fixed in this PR:
https://github.com/intive/patronage2023-js/pull/116

apps/web/lib/types.ts
```diff
@@ -22,6 +22,13 @@ export interface Transaction {
   creator: Creator;
 }
 
+export interface BudgetUser {
+  id: string;
+  avatar: string;
+  firstName: string;
+  lastName: string;
+}
+
```
bdeluga, line 31: Thank you, for making this file usefull, that's my precious baby 💖

bdeluga (approved):
LGTM 🚀, as of issue you mentioned, it would be best if backend changed those weird `src` url's with `""` so we can render default avatar, when src is `falsy`.

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
```diff
@@ -0,0 +1,83 @@
...
+    font-weight: 600;
+  }
+`;
+
+const StyledUser = styled.div`
+  padding: 3px;
+`;
+
+const PeopleInBudget = ({ users }: PeopleInBudgetProps) => {
+  const range = users.length > 4 ? 3 : users.length;
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx:48]: @wyola You don't need this - you can just do `users.slice(0,3)` and if the array has less than 3 members, it will just give you the whole array anyway :).

packages/ui/Avatar/index.tsx
```diff
@@ -21,19 +22,30 @@ const Image = styled.img<AvatarProps>`
...
+      username = "user",
+      src,
+      alt,
+      className,
+      outlined,
+      // other components, like Tooltip, might add
+      // custom props to extend functionality of this component
+      ...otherProps
```
monika-klokosinska-intive, line 35: @wyola this is potentially dangerous - we don't know what props are being passed here, which makes it harder to maintain and debug. Could you just list all the props the tooltip needs and pass them one by one? The prop list will get bloated, but it is the lesser of the two evils.

monika-klokosinska-intive (commented):
@wyola two minor code suggestions from me.
Another thing: could you connect with @PawelPazderski and see if you can implement his `default avatar` behaviour. Broken avatar right displays a gigantic alt which ruins the visuals - and ruins your hard work as well.

It is fine to address the default avatar in a separate PR if it is more convenient to merge it as it for now! Demo is coming, so let's be super agile.

Noshi96 (approved):
LGTM 🚀 🎉

zuzannapaw (approved):
Nicely applied changes and great work! 🚀

mpiwonski (approved):
Good job!

Vectari (approved):

AzulloCyll:
Please merga actual main to get rid of app crashes.

aniasgit (approved):
Looks good to me 👍

PawelPazderski (approved):
LGTM 🚀

apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx
Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/PeopleInBudget.tsx]: @monika-klokosinska-intive fixed

packages/ui/Avatar/index.tsx
monika-klokosinska-intive: @wyola alright, I will let it slide this time since it does include integrating with radix.

monika-klokosinska-intive (approved):

monika-klokosinska-intive (approved):

PR #117, opened 2023-05-24
feat: budget - sorting in columns [UI] #P23-225

mpiwonski:
Task: [Jira](https://tracker.intive.com/jira/browse/PATRO23-225)

### How it works

A User is able to click sorting icon next to column title, which by default remains inactive (light gray color), and once they do that, sorting has been activated, that is indicated by color change (dark gray color) and upward arrow is displayed (ascending). When a user clicks sorting icon again, sorting is switched to descending mode, causing an arrow icon to go downwards. Once a user clicks any sorting icon, they will just be able to go from one sorting mode to another.

Default state: sorting ascending by name (description) - included in request body. Sorting by date has been provided by backend.

### To be done

Still, there's no backend endpoint to make this function work, so there's just UI change. You may take a look at the console to see how the state is being changed and what would be sent to backend in case the connection was established.

### DIFF
--- a/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
@@ -57,24 +57,16 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
 
   const setSorting = (column: string) => {
     setSortDescriptors((previousSortDescriptors) => {
-      if (
-        previousSortDescriptors.find((element) => element.columnName === column)
-      ) {
-        return [
-          ...previousSortDescriptors.filter(
-            (element) => element.columnName !== column
-          ),
-          {
-            columnName: column,
-            sortAscending: !previousSortDescriptors.find(
-              (element) => element.columnName === column
-            )?.sortAscending,
-          },
-        ];
-      }
       return [
-        ...previousSortDescriptors,
-        { columnName: column, sortAscending: true },
+        ...previousSortDescriptors.filter(
+          (element) => element.columnName !== column
+        ),
+        {
+          columnName: column,
+          sortAscending: !previousSortDescriptors.find(
+            (element) => element.columnName === column
+          )?.sortAscending,
+        },
       ];
     });
   };
@@ -122,7 +114,6 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
     ],
 
     queryFn: async () => {
-      console.log(sortDescriptors);
       return fetch(
         `${env.NEXT_PUBLIC_API_URL}budgets/${budget.id}/transactions`,
         {
--- a/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
@@ -46,12 +46,38 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
     "Income" | "Expense" | null
   >(null);
   const [searchTransactionByName, setSearchTransactionByName] = useState("");
+  const [sortDescriptors, setSortDescriptors] = useState([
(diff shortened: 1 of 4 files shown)


apps/web/app/(navigation)/budgets/[id]/TransactionsTable.tsx
```diff
@@ -201,10 +211,26 @@ export const TransactionsTable = ({
                 {column.key !== "editColumn" && (
                   <button onClick={() => setSorting(column.key)}>
                     <Icon
-                      icon="sort"
+                      icon={"sort"}
```
Comment (aniasgit) [inline apps/web/app/(navigation)/budgets/[id]/TransactionsTable.tsx:214]: if we are passing a string these braces are not needed

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -46,13 +46,39 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
...
   const debouncedSearch = useDebounce(searchTransactionByName, 500);
   const { t, dict } = useTranslate("BudgetsPage");
-  const setSorting = (column: string) => console.log(column);
   const { data: session } = useSession();
   const categoryFilterState = useAtomValue(categoryFilterAtom);
   const pageSize = parseInt(getPageSizeValue);
 
+  const setSorting = (column: string) => {
+    setSortDescriptors((previousSortDescriptors) => {
+      if (
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:60]: @mpiwonski I think this if is not necessary.

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -46,13 +46,39 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
...
+        previousSortDescriptors.find((element) => element.columnName === column)
+      ) {
+        return [
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:63]: @mpiwonski this `return` should also work in case of the element not being inside the `previousSortDescriptors`.

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -92,9 +118,11 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
       categoryFilterState,
       transactionType,
       debouncedSearch,
+      sortDescriptors,
     ],
 
     queryFn: async () => {
+      console.log(sortDescriptors);
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:125]: @mpiwonski please remove the console.log!

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (mpiwonski) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:63]: If the element is not present in previousSortDescriptors, it means it has not been activated yet, so when a user clicks sorting button for the first time, it will become active, thus an object with column name and sorting type is added.

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -46,13 +46,39 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
...
+      if (
+        previousSortDescriptors.find((element) => element.columnName === column)
+      ) {
+        return [
+          ...previousSortDescriptors.filter(
+            (element) => element.columnName !== column
+          ),
+          {
+            columnName: column,
+            sortAscending: !previousSortDescriptors.find(
```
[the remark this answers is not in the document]
Comment (mpiwonski) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:69]: I'm not really sure where undefined could come from, if, first, we check whether particular column already exists in state?

monika-klokosinska-intive (approved):

wyola (approved):
LGTM 🚀

PR #118, opened 2023-05-25
fix: people in budget #P23-219

wyola:


monika-klokosinska-intive:
Discussed with @wyola privately and approved by our leadership (me).

PR #119, opened 2023-05-26
refactor: top section budget details #P23-357 

wyola:
## Layout of the top section of the budget's details page
task: https://tracker.intive.com/jira/browse/PATRO23-357

Let me know what you think about the layout of the top part, try resizing the page down to 375px.

It is best if you test it by logging in with the same credentials as for "People in budget":
https://intive-apu8934.slack.com/archives/C04R6LM4JSV/p1684873435505349?thread_ts=1684873404.585919&cid=C04R6LM4JSV

Task is related to this part:

image

### Notes:
- it has to look good both for short and long budget names/descriptions, so there is no satisfying every possibility 

### What changed:
- extra icon for adding the budget to favorite
- icons move to dropdown menu on smaller screens
- layout HTML structure, because of the "people in budget task"
- suspense was updated accordingly
- small refactor of the `Favourite` component - when a different budget was loaded using side nav menu the value of isFavourite was not updating, hence:

```
  useEffect(() => {
    setIsFavourite(isFav);
  }, [budgetId, isFav]);
```
- I changed `TransactionDropdownMenu` to have more generic name `DropdownMenu`, since it is now used not only in `TransactionsTable` but also in budget's details page mobile view

### DIFF
--- a/.github/workflows/lint-pr.yml
+++ b/.github/workflows/lint-pr.yml
@@ -11,7 +11,6 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-latest
-
     name: Lint
     steps:
       - uses: actions/checkout@v3
@@ -26,7 +25,6 @@ jobs:
 
   prettier:
     runs-on: ubuntu-latest
-
     name: Prettier
     steps:
       - uses: actions/checkout@v3
@@ -38,3 +36,17 @@ jobs:
           node-version: 16
       - run: npm ci
       - run: npm run prettier
+
+  dict-check:
+    runs-on: ubuntu-latest
+    name: Check dictionary
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Run dictionary checker
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm ci
+      - run: npm run dict-check
--- a/apps/storybook/stories/TransactionDropdownMenu.stories.tsx
+++ b/apps/storybook/stories/TransactionDropdownMenu.stories.tsx
@@ -63,9 +63,7 @@ export default {
   },
 } as ComponentMeta<typeof DropdownMenu>;
 
-const Template: ComponentStory<typeof DropdownMenu> = ({
-  ...args
-}) => (
+const Template: ComponentStory<typeof DropdownMenu> = ({ ...args }) => (
   <DropdownMenuWrapper>
     <DropdownMenu {...args} />
   </DropdownMenuWrapper>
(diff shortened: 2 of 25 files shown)


Alessiass (approved):

Noshi96 (commented):
Looks great! I have noticed that the tooltip doesn't open on mobile. What do you think, is there a way to make it open when clicked? If so, we could also apply this to long budget titles that we can't check on mobile until we get to the edit modal.

zuzannapaw (approved):
I agree with @Noshi96 idea, however it looks superb ✨ Good job

wyola:
> Looks great! I have noticed that the tooltip doesn't open on mobile. What do you think, is there a way to make it open when clicked? If so, we could also apply this to long budget titles that we can't check on mobile until we get to the edit modal.

@Noshi96 I semi-agree 😅 there probably should be some functionality for it on mobile, but tooltips are not displayed on mobiles. Then it would have to be maybe something else, like a `Popover`, then the `Avatar` should be a button, then maybe it should lead to user's account's details, which we don't have etc etc... We would be starting an avalanche here. Also, radix's tooltip does not really have an option to add this functionality, and the `people in budget` task was about hovering, and this task is about the layout.

Noshi96:
> @Noshi96 I semi-agree 😅 there probably should be some functionality for it on mobile, but tooltips are not displayed on mobiles. Then it would have to be maybe something else, like a `Popover`, then the `Avatar` should be a button, then maybe it should lead to user's account's details, which we don't have etc etc... We would be starting an avalanche here. Also, radix's tooltip does not really have an option to add this functionality, and the `people in budget` task was about hovering, and this task is about the layout.

@wyola it's ok, we don't want an avalanche 😅  I thought there might be an easy way to apply this functionality. And you're right that it's beyond the scope of people on a budget task.

PawelPazderski (approved):
Looks and works really nice 🚀 Maybe we could make the budget icon even smaller at mobile view and give more space for information, but I believe this will look better when the side nav stops stealing the space. One small issue I ran into while changing the window size while dropdown open:

image

Other than that, good job.

wyola:
> Looks and works really nice 🚀 Maybe we could make the budget icon even smaller at mobile view and give more space for information, but I believe this will look better when the side nav stops stealing the space. One small issue I ran into while changing the window size while dropdown open:
> 
> image
> 
> Other than that, good job.

thanks, I fixed it 🔨 nice drawings btw 😄

Noshi96 (approved):
Oh, I discovered that on mobile, when I touch and hold a specific avatar, the alternate text of that avatar is displayed, which solves my question about which avatar belongs to the user. Other than that, great job! ✨

Vectari (approved):

monika-klokosinska-intive (approved):

PR #120, opened 2023-05-27
refactor: transaction table

wlechowicz:
Initial refactor of the transaction table.
Step 1 in trying to make table controller more generic and reusable is making the table more generic.


bdeluga (approved):
LGTM, nice idea to create hook for locale date tho 🚀

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -128,7 +127,10 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
           },
         }
       )
-        .then((res) => fixFetchedData(res))
+        .then((res) => {
+          setTotalPages(Math.ceil(res.totalCount / pageSize));
```
Comment (wyola) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:131]: Didn't notice this in the previous PR, but it is possible for the user to set the value manually in local storage to 0, and we will have division by zero here. Nothing major I guess, since user should not aim to break the page for themselves xD but it would be probably nice to fix it.

Comment (wyola) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:131]: Didn't notice this in the previous PR, but it is possible for the user to set the value manually in local storage to 0, and we will have division by zero here. Nothing major I guess, since user should not aim to break the page for themselves xD but it would be probably nice to fix it.

wyola (approved):
just a minor thing I noticed

apps/web/lib/dictionary.ts
```diff
@@ -1,4 +1,10 @@
 const dictionary = {
+  Common: {
+    days: {
+      yesterday: { en: "Yesterday", pl: "wczoraj", fr: "hier" },
+      today: { en: "Today", pl: "dzisiaj", fr: "aujourd'hui" },
+    },
+  },
```
zuzannapaw, line 7: Was it intended to start day's names with a lowercase letters in polish and french?

zuzannapaw (approved):
Nice changes, also great idea with setting more generic name for transactions table suspense (now simply `TableSupsense`). 🚀  Only one question from me 👀

apps/web/lib/dictionary.ts
wyola: @zuzannapaw I think this is how the dayjs library generates it for the week days, so those are probably just adjusted to that

monika-klokosinska-intive (approved):

apps/web/app/(navigation)/budgets/[id]/TransactionsTable.tsx
```diff
@@ -201,49 +180,46 @@ export const TransactionsTable = ({
               switch (props.column.key) {
                 case "date":
                   const value = props.groupKey[props.groupIndex];
-                  return <>{getDayName(value)}</>;
+                  return <>{getLocaleDateString(value)}</>;
```
Comment (Noshi96) [inline apps/web/app/(navigation)/budgets/[id]/TransactionsTable.tsx:183]: @wlechowicz can we just return a string here instead of a fragment, or does it not matter at all?
```suggestion
                  return getLocaleDateString(value);
```

Noshi96 (approved):
Works great! 🚀 One question from me:

Alessiass (approved):

Vectari (approved):

apps/web/lib/dictionary.ts
wlechowicz: I didn't change this, just moved it about. While day names in English start with a capital, I don't think "today" and "yesterday" should. In UX terms, I think it should be the same for all languages - capitalized. But that's UI layer, so should be done via CSS and they probably should be all lowercase in dict.

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (wlechowicz) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:131]: Yes, probably `setTotalPages(pageSize > 0 Math.ceil(res.totalCount / pageSize) : 1);` but I was just moving code around for refactoring purposes.

PR #121, opened 2023-05-27
fix: email resets after pressing enter in form #P23-368

Noshi96:
### Why does it happen ❓ 
In forms, clicking enter triggers a button, which is used to trigger submit.
In our case, the icons inside the inputs are buttons, so clicking enter would trigger the first button in the queue (1. Clear input button, 2. Show/hide password button, 3. Submit button). If we remove the clear function, the next button is triggered.

### How can this be prevented ❓ 
```tsx
<button type="button">Show</button>
```

Setting `type="button"` in buttons makes them ignored when enter is clicked.

### DIFF
--- a/packages/ui/Input/index.tsx
+++ b/packages/ui/Input/index.tsx
@@ -46,6 +46,7 @@ export const Input = forwardRef(
       if (type === "password") {
         return (
           <StyledIcon
+            type="button"
             onClick={(event) => {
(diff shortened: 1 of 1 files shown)


wlechowicz (approved):

bdeluga (approved):
That's the PR i love to see ! 🚀

PR #122, opened 2023-05-27
feat: hamburger menu for side nav mobile view #P23-366

PawelPazderski:
## Feature improving mobile experience by displaying a hamburger menu that triggers side navigation. 
Ticket: https://tracker.intive.com/jira/browse/PATRO23-366

For better user experience there is a function that closes side nav when clicking outside of it provided with the menu. 
The menu would also close when the path changes.

Side nav is still displayed same as before at tablet breakpoint & higher.

As we don't have designs for this I decided to add slight background color to the side nav, IMO that stands out better as we have white background underneath.

I used the library below to deliver nicely animated and easy to configure & use hamburger menu component:
https://github.com/luukdv/hamburger-react

Few thigs also in PR:
- neither side nav nor the hamburger menu are displayed when user is not logged in 
- some slight changes to pagination styling at mobile view


wlechowicz:
Looks nice and works well! I spotted some issues with clicking stuff on mobile, but it's the same on main, so it's not due to your work. Please verify that the issues you're listing aren't happening on main already.

wlechowicz:
Finish up your open issues and mark as ready for review, then team should review the code (me included).

apps/web/app/(navigation)/Nav.tsx
```diff
@@ -31,10 +36,33 @@ const AvatarStyled = styled(Avatar)`
...
+  const toggleMenu = () => {
+    setOpen((prev) => !prev);
+  };
+
   return (
     <NavBar>
+      {data && (
+        <BurgerWrapper>
+          <Hamburger
+            label="Show menu"
```
wyola, line 58: use dictionary please

apps/web/app/(navigation)/Nav.tsx
```diff
@@ -31,10 +36,33 @@ const AvatarStyled = styled(Avatar)`
...
+            color="#FFF"
```
wyola, line 59: use theme please

apps/web/app/(navigation)/layout.tsx
```diff
@@ -20,20 +26,70 @@ const Main = styled.main`
...
+  ${device.tablet} {
+    padding-left: 94px;
+  }
+`;
+
+const ContentNoUser = styled.div`
+  flex-grow: 1;
+  width: 100%;
+  padding-left: 25px;
+`;
```
wyola, line 42: you could simplify and get rid of the code repetition like below:

```
const ContentNoUser = styled.div`
  flex-grow: 1;
  width: 100%;
  padding-left: 25px;
`;

const ContentUser = styled(ContentNoUser)`
  ${device.tablet} {
    padding-left: 94px;
  }
`;
```

apps/web/app/(navigation)/layout.tsx
```diff
@@ -20,20 +26,70 @@ const Main = styled.main`
...
+  ${device.tablet} {
+    display: none;
+  }
+`;
+
+const SideNavDesktop = styled.div`
+  @media (max-width: 767px) {
+    display: none;
+  }
 `;
```
wyola, line 54: why hardcoding it? if `tablet` size in media queries would change, then you have to remember to change this hardcoded value

```suggestion
const SideNavDesktop = styled.div`
  display: none;
  ${device.tablet} {
    display: block;
  }
`;
```

wyola (commented):
Great addition to the app! ⭐

apps/web/app/(navigation)/Nav.tsx
PawelPazderski: ok, done

apps/web/app/(navigation)/Nav.tsx
PawelPazderski: done, thanks

apps/web/app/(navigation)/layout.tsx
PawelPazderski: good idea, thanks

apps/web/app/(navigation)/layout.tsx
PawelPazderski: done, thanks a lot for the review 🚀

AzulloCyll (approved):
Works and looks great! 🚀

wyola (approved):

apps/web/app/(navigation)/Nav.tsx
```diff
@@ -1,9 +1,16 @@
...
+import { Turn as Hamburger } from "hamburger-react";
+import styled from "styled-components";
+import { device } from "lib/media-queries";
 import { LanguageSelector } from "./LanguageSelector";
 import { Logo, Avatar } from "ui";
+import { mobileMenuAtom } from "store";
+import { useAtom } from "jotai";
+import { useContext } from "react";
+import { ThemeContext } from "styled-components";
+import { useTranslate } from "lib/hooks";
```
zuzannapaw, line 13: I suggest to place external improts first, and after that our internal imports ✔️

apps/web/components/Pagination/Pagination.styled.tsx
```diff
@@ -21,12 +25,13 @@ export const RowsPerPageContainerStyled = styled.div`
 
 export const ListStyled = styled.ul`
   display: flex;
-  gap: 8px;
-  flex-wrap: wrap;
-
+  gap: 0;
```
zuzannapaw, line 28: `gap: 0` can be deleted ☺️

zuzannapaw (approved):
Great addition! Well done 🚀Small changes from me.

I have as well one suggestion, I'm not sure if the new background color of side nav on mobile is needed, I don't feel like this shade fits well. It still looks good with previous gray color `#F7F7F7`. But that's just my opinion, I'd be happy to hear what do you think about it!

image

Still approved! 😎

apps/web/app/(navigation)/layout.tsx
```diff
@@ -20,20 +26,71 @@ const Main = styled.main`
...
+      setSideOpen(false);
+    };
+
+    const mainElement = document.querySelector("main") as HTMLElement;
+
+    mainElement.addEventListener("mousedown", handleClickOutside);
+    return () => {
+      mainElement.removeEventListener("mousedown", handleClickOutside);
+    };
+  }, [isSideOpen, setSideOpen]);
```
bdeluga, line 74: This looks like a perfect example of what hook could looks like, no adjustments needed. 

Now click outside works only for `main`, what about when we click on header, should we hide or not ?

bdeluga (approved):
Nice job ! 🚀

apps/web/app/(navigation)/layout.tsx
PawelPazderski: I added the listener to main, not the document, because when added it on document made it impossible to close the menu by clicking the X icon on nav. The X icon is in the nav and it's natural to use it there so I figured it would be ok to do so, my main goal was to add this functionality to main content in the first place.

apps/web/components/Pagination/Pagination.styled.tsx
PawelPazderski: right

apps/web/app/(navigation)/Nav.tsx
PawelPazderski: yeah you're right I keep forgetting that

PawelPazderski:
> Great addition! Well done 🚀Small changes from me.
> 
> I have as well one suggestion, I'm not sure if the new background color of side nav on mobile is needed, I don't feel like this shade fits well. It still looks good with previous gray color `#F7F7F7`. But that's just my opinion, I'd be happy to hear what do you think about it!
> 
> This is how it would look like with previous color: image
> 
> Still approved! 😎

Maybe the problem is that I can barely see the difference between white and this #F7F7F7 on my monitor, looks almost the same so I decided to add some color to make it stand out. I do like it this way but it will be ok if we decide to stay on that grey background as well.

Alessiass (approved):
Looks nice and works nice - I really enjoy the animation

Noshi96 (approved):
The hamburger works very well. I see that you centered the pagination, and it looks much better on mobile now. For me, the old navigation bar background color was more consistent with the rest of the layout. Nevertheless, good job! 👍

PawelPazderski:
> The hamburger works very well. I see that you centered the pagination, and it looks much better on mobile now. For me, the old navigation bar background color was more consistent with the rest of the layout. Nevertheless, good job! 👍

You may be right, Zuzanna said same thing, I wanted it to stand out more and - what is more - my monitor I display the browser with has some - let's say - issue with greys, it's old 💩 to be honest - below some value, like about 5% I can't see difference between white and such color. I will bring the color back as I'm getting more and more convinced after your comment.. thanks

Vectari (approved):

monika-klokosinska-intive (approved):

apps/web/app/(navigation)/layout.tsx
wlechowicz: This is usually solved with a useClickOutside custom hook that does what you did here, but it accepts a ref as a param so that you can attach a ref to any element and pass that ref to the hook, so it will work for clicks outside a specific element in the DOM. Such a hook is then very universal and it's possible to nest them inside the same branch of the DOM tree.
You don't have to select elements from the document via selectors because you have a ref.

wlechowicz (approved):
Possible improvement is the useClickOutside custom hook, other than that - LGTM

PR #123, opened 2023-05-28
feat: sign-out #P23-19

AzulloCyll:
Task: https://tracker.intive.com/jira/secure/RapidBoard.jspa?rapidView=1661&projectKey=PATRO23&view=detail&selectedIssue=PATRO23-19&sprint=5402

DONE:
- a small select menu will appear when user clicked on avatar.
- one of options enables ability to sign out from InBudget application.

TODO:
Add toast maybe.
Wait for https://github.com/intive/patronage2023-js/pull/122 by @PawelPazderski to be done, to check possible conflicts and resolve them


apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const { data } = useSession();
+  const { t, dict } = useTranslate("MainPage");
+
+  const menuHandler = (value: string) => {
+    if (value === "sign-out") {
+      router.push("/");
+      signOut();
+    }
+    router.push("/");
+  };
```
wyola, line 32: You don't need `router.push("/")` here, because it is handled by `next-auth` here:

https://github.com/intive/patronage2023-js/blob/5609f071eb22fb050e9fcf7ca8a1b84d67e84e6c/apps/web/pages/api/auth/%5B...nextauth%5D.ts#L83

if you also remove the `home` option then you don't need both `router.push("/")`

```suggestion
  const menuHandler = (value: string) => {
    if (value === "sign-out") {
      signOut();
    }
  };

**Review by wyola — COMMENTED** [review]
Good idea with the `sign out` option, it was missing in the app! Few remarks from me

File: apps/web/app/(navigation)/MainMenu.tsx [hunk 2]
```diff
@@ -0,0 +1,74 @@
...
+export const MainMenu = () => {
+  const { hasScrollbar } = useHasScrollBar();
+  const router = useRouter();
+  const { data } = useSession();
+  const { t, dict } = useTranslate("MainPage");
+
+  const menuHandler = (value: string) => {
+    if (value === "sign-out") {
+      router.push("/");
+      signOut();
```
bdeluga, line 29: SignOut, "pushes" you to "/" on it's own, dont need to `router.push("/")`

bdeluga (commented):
I feel like this could simply be a dropdown, which renders `Link` with `href="/"`  and a button with `signOut` on click. @zuzannapaw already provided us with TransactionDropdownMenu. You could just make it more generic to render main item `avatar` in this case and just toss said components to it's props.

@monika-klokosinska-intive  @wlechowicz  what do you guys think ? 

Great work nontheless, it is working as expected 🚀

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+            <SelectContentStyled
+              position="popper"
+              align="center"
+              sideOffset={5}>
+              <Select.Viewport>
+                {items.map((item) => (
+                  <SelectItemStyled value={item.value} key={item.value}>
+                    <Select.ItemText>{item.text}</Select.ItemText>
+                  </SelectItemStyled>
+                ))}
```
zuzannapaw, line 65: You could use object's destructuring here 
```suggestion
                {items.map(({value, text}) => (
                  <SelectItemStyled value={value} key={value}>
                    <Select.ItemText>{text}</Select.ItemText>
                  </SelectItemStyled>
                ))}
```

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+  SelectItemStyled,
+  SelectTriggerStyled,
+} from "./LanguageSelectorStyled";
+import styled from "styled-components";
+
+import { useHasScrollBar } from "lib/hooks/useHasScrollBar";
+import { Avatar } from "ui";
+import { useRouter } from "next/navigation";
+import { signOut, useSession } from "next-auth/react";
+import { useTranslate } from "lib/hooks";
```
zuzannapaw, line 13: I suggest to place external imports first, and after that our internal imports

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+      signOut();
+    }
+    router.push("/");
+  };
+
+  const items = [
+    {
+      value: "home",
+      text: t(dict.home),
+    },
```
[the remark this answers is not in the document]
zuzannapaw: I agree with @wyola, I believe it's reduntant

zuzannapaw (commented):
What a great improvement be to be able to log out from any page now! ✨ 

Only few things from me.  @bdeluga and @wyola proposed very apt suggestion about `router.push()` 👀 

Well done 🚀 
PS. As we said before, I let myself commit from your branch one amendment of french text 🐸

apps/web/app/(navigation)/MainMenu.tsx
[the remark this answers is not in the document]
AzulloCyll: Home was added, like a placeholder. To show that's in that select we can have more than one options. Also select with only one option looks radicolous. :P

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+  ];
+
+  return (
+    <>
+      {data && (
+        <Select.Root onValueChange={(value) => menuHandler(value)}>
+          <SelectTriggerStyled>
+            <Select.Value>
+              <AvatarStyled src="avatars/3.svg" outlined />
```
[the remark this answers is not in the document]
AzulloCyll: Ooo. Thank you to paying attention to it. I'am restoring it.

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+    {
+      value: "sign-out",
+      text: t(dict.signOut),
+    },
+  ];
+
+  return (
+    <>
+      {data && (
+        <Select.Root onValueChange={(value) => menuHandler(value)}>
```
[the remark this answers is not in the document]
AzulloCyll: I need this value on handler, because I decided not to delete "Home" option flrom select menu

apps/web/app/(navigation)/MainMenu.tsx
AzulloCyll: You right, I don't need router.push on sign-out route, but i'm leaving another one to make "Home button work"

apps/web/app/(navigation)/MainMenu.tsx
AzulloCyll: I used select, brecause i've styles for other select written already, and I tried to make both select's looks similar.

apps/web/app/(navigation)/MainMenu.tsx
AzulloCyll: That's good sugestion :)

apps/web/app/(navigation)/MainMenu.tsx
AzulloCyll: Ok, I'll do it.

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const router = useRouter();
+  const { data } = useSession();
+  const { t, dict } = useTranslate("MainPage");
+
+  const menuHandler = (value: string) => {
+    if (value === "sign-out") {
+      signOut();
+    }
+    router.push("/");
+  };
```
Alessiass, line 32:
```suggestion
  const menuHandler = (value: string) => {
    value === "sign-out" ? signOut() : "";
  };
  };
```

The signOut function already has redirect function built in thus push will never fire. This can be simplified since you cannot access this menu if you are not logged in already.

Alessiass (changes requested):

Alessiass (approved):

apps/web/app/(navigation)/MainMenu.tsx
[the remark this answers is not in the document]
wyola: @AzulloCyll to me personally it looks more ridiculous to have "home" there ;D I don't see anything strange in having there just the sign out. But that's just my personal preference, keep it as you wish 🤗

apps/web/app/(navigation)/MainMenu.tsx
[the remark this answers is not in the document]
wyola: @AzulloCyll with suggested approach the `value` will still be passed to the `menuHandler` :)

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -0,0 +1,70 @@
...
+              <Select.Viewport>
+                {items.map(({ value, text }) => (
+                  <SelectItemStyled value={value} key={value}>
+                    <Select.ItemText>{text}</Select.ItemText>
+                  </SelectItemStyled>
+                ))}
+              </Select.Viewport>
+            </SelectContentStyled>
+          </Select.Portal>
+        </Select.Root>
```
Noshi96, line 66: Do we need radix select here? Isn't it better to use radix dropdown? What do you think?

Noshi96 (commented):
I also agree with @zuzannapaw that we need some sort of indicator to show the user that they can interact with the avatar. Other than that, good job as the logout works as it should 🚀

apps/web/app/(navigation)/MainMenu.tsx
AzulloCyll: I really really don't know. I've used select because I'v had all the styles already written, so ive decided to do it on select. 🤔

AzulloCyll:
> I also agree with @zuzannapaw that we need some sort of indicator to show the user that they can interact with the avatar. Other than that, good job as the logout works as it should 🚀

I've added indicator :)

apps/web/app/(navigation)/MainMenu.tsx
AzulloCyll: I've added arrow indicator also.

PawelPazderski (commented):
Looks nice but ran into unexpected behaviour, as a user I expect to log out from any place in the app and it should redirect me to login page, meanwhile it doesn't happen so everywhere 👇 

https://github.com/intive/patronage2023-js/assets/72406958/4fe957c9-b603-4b30-ab43-83da527ee424

AzulloCyll:
> Looks nice but ran into unexpected behaviour, as a user I expect to log out from any place in the app and it should redirect me to login page, meanwhile it doesn't happen so everywhere 👇

I know about it. 👌🏻 
In PR with refiment from @bdeluga there will be no left menu for unlogged user, so we will always be redirected to sign-in page. So if we merge it the behavior will be correct :)

It's beahavior is as is because that path is no protected by middleware. (That's was a cause why I've added router.push("/") originally too).

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -0,0 +1,84 @@
...
+      value: "sign-out",
+      text: t(dict.signOut),
+    },
+  ];
+
+  return (
+    <>
+      {data && (
+        <Select.Root
+          onValueChange={(value) => menuHandler(value)}
```
wyola, line 56: I am pointing that one out again ;-) `menuHandler` is already a function that takes `value`

```suggestion
          onValueChange={menuHandler}
```
but I worked out that there is an issue anyway. With the way you have it right now, redirecting to `home` works only the first time. After that, `select` "remembers" the previously chosen value and does not call `onValueChange`.

check this with the current code that you have: 
- login
- stay on the home page or go the budget page
- select `home` from your select
- page reloads correctly
- go to other budget page or stay on home and select home again - page does not reload anymore

How to fix it? add `value=""` in `Select.Root`
```
        <Select.Root
          value=""
          onValueChange={menuHandler}
```
... and please include the suggested change, since it is really the same :)

wyola (changes requested):
found a bug 🐛

apps/web/app/(navigation)/MainMenu.tsx
AzulloCyll: Oooo! Youre right, that is wierd. I'll fix it.

apps/web/app/(navigation)/MainMenu.tsx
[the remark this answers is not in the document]
AzulloCyll: fixed

monika-klokosinska-intive (approved):

wyola (approved):

PR #124, opened 2023-05-28
fix: budget list search #P23-375

Vectari:
# Hello! 👋
Task [Fix Budget List Search](https://tracker.intive.com/jira/browse/PATRO23-375)

## What fixes:
1. Now when you type something in search bar and change tab (e.g. Reports) and open Budgets again, search bar will be empty and budget list will show all budgets.
2. When you go to specyfic budget after search and then open again budget list, search bar again will be empty and budget list will show all budgets.

Thanks to @Noshi96 for a hint how to fix it in a comment on Slack :)

### DIFF
--- a/apps/web/app/(navigation)/SideNavigationBar.tsx
+++ b/apps/web/app/(navigation)/SideNavigationBar.tsx
@@ -138,7 +138,7 @@ export default function SideNav() {
       },
       resetSearch: () => {
         setSearchValue("");
-      }
+      },
     },
     navigationList: (
       <NavList
--- a/apps/web/app/(navigation)/SideNavigationBar.tsx
+++ b/apps/web/app/(navigation)/SideNavigationBar.tsx
@@ -237,7 +237,7 @@ export default function SideNav() {
         isNavListItemClicked={isNavListItemClicked}
         resetIsNavListItemClicked={resetIsNavListItemClicked}
         refetchBudgetsFunction={refetch}
-        resetSearch={setSearchValue}
+        resetSearch={() => setSearchValue("")}
       />
       <>
         {isCreateNewBudgetModalVisible && (
--- a/apps/web/app/(navigation)/SideNavigationBar.tsx
+++ b/apps/web/app/(navigation)/SideNavigationBar.tsx
@@ -136,6 +136,9 @@ export default function SideNav() {
       onChange: (value: string) => {
         setSearchValue(value);
       },
+      resetSearch: () => {
+        setSearchValue("");
+      }
     },
     navigationList: (
       <NavList
@@ -237,6 +240,7 @@ export default function SideNav() {
         isNavListItemClicked={isNavListItemClicked}
         resetIsNavListItemClicked={resetIsNavListItemClicked}
         refetchBudgetsFunction={refetch}
+        resetSearch={setSearchValue}
       />
       <>
         {isCreateNewBudgetModalVisible && (
--- a/apps/web/app/(navigation)/SideNavigationBar.tsx
+++ b/apps/web/app/(navigation)/SideNavigationBar.tsx
@@ -136,9 +136,6 @@ export default function SideNav() {
       onChange: (value: string) => {
         setSearchValue(value);
       },
-      resetSearch: () => {
-        setSearchValue("");
-      },
     },
     navigationList: (
       <NavList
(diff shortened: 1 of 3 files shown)


packages/ui/SideNavigationBar/index.tsx
```diff
@@ -72,6 +75,7 @@ export const SideNavigationBar = ({
     setSubMenuId(id);
     resetIsNavListItemClicked();
     setIsSubMenuShown(true);
+    resetSearch();
   };
```
Noshi96, line 79: Here, when you reset it after opening the submenu, you force an additional request and refresh the list, but the search state is the same (empty string). Setting `resetSearch();` in `hideSubMenu()` I think is enough. Or is there some other use case that I don't see?
```suggestion
    setSubMenuId(id);
    resetIsNavListItemClicked();
    setIsSubMenuShown(true);
  };
```

Noshi96 (changes requested):
Works well! Minor suggestions below

packages/ui/SideNavigationBar/index.tsx
Vectari: I use it to refresh budget list when you are in specific budget after search. Number 2 in my PR describe.

apps/web/app/(navigation)/SideNavigationBar.tsx
```diff
@@ -136,6 +136,9 @@ export default function SideNav() {
       onChange: (value: string) => {
         setSearchValue(value);
       },
+      resetSearch: () => {
+        setSearchValue("");
+      },
     },
```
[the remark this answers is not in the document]
Vectari: You are right. Fixed :)

packages/ui/SideNavigationBar/index.tsx
Noshi96: Okay now I see, my mistake :)
I'm still wondering, because when you reopen after changing the budget for a second, you will see the old searched budget 🤔

Vectari: Probably debounce. I’ll try to do something with it

Noshi96: Previously, when the cleanup worked, there was a useEffect in the SubMenu/index.tsx file that called onChange on the input with an empty string when the component was unmounted. I'm not sure if this was the best solution:
```tsx
  useEffect(() => {
    return () => {
      searchInput?.onChange?.("");
    };
  }, []);
```

Vectari: Yes but with it there was ESlint warning and probably (I don't really remember) it don't work in case described in number 2 in PR description.

Noshi96: It works, but you are right about the ESlint warning caused by the empty dependency array in useEffect.

apps/web/app/(navigation)/SideNavigationBar.tsx
```diff
@@ -237,6 +237,7 @@ export default function SideNav() {
         isNavListItemClicked={isNavListItemClicked}
         resetIsNavListItemClicked={resetIsNavListItemClicked}
         refetchBudgetsFunction={refetch}
+        resetSearch={setSearchValue}
```
wyola, line 240: You are calling a function without a parameter, which causes setting this state to `undefined`, while it should be of a `string` type. Doesn't seem to break anything, but let's be on the safe side and have the proper type 🤓 

```suggestion
        resetSearch={() => searchValue("")}
```

search1

after closing the side nav, the state is `undefined` ⬇️ 

search2

wyola (approved):
just one minor thing from me, other than that works good!

AzulloCyll (approved):

zuzannapaw (approved):
Works smoothly! 🚀

Noshi96 (approved):

Alessiass (approved):

apps/web/app/(navigation)/SideNavigationBar.tsx
Vectari: Great idea. Fixed 🙂

PawelPazderski (approved):
Works as it should, nice. Only one tiny little thing it's not some major problem just cought my eye - when we search the budget - go into the searched one and then open the list again we can see the filtered list for a slight moment before the list loads again, other than that 🎸 🚀

monika-klokosinska-intive (approved):

PR #126, opened 2023-05-28
fix: credentials error #P23-376

wyola:
Tiny fix for the situation when you type in the wrong credentials and the error does not display, instead, the page reloads.

I don't know why this got removed in some previous PR:
`redirect: false,`

### DIFF
--- a/apps/web/app/(regflow)/sign-in/page.tsx
+++ b/apps/web/app/(regflow)/sign-in/page.tsx
@@ -68,6 +68,7 @@ export default function SignInPage() {
         signIn("credentials", {
           email: values.email,
           password: values.password,
+          redirect: false,
         }).then((res) => {
           if (res!.ok) {
             router.push("/");
--- a/apps/web/app/(regflow)/sign-in/page.tsx
+++ b/apps/web/app/(regflow)/sign-in/page.tsx
@@ -34,7 +34,7 @@ const FieldsWrapper = styled.div`
 const ErrorWrapper = styled.div`
   position: absolute;
   width: 100%;
-  top: 1rem;
+  top: -1rem;
   left: 50%;
   transform: translateX(-50%);
   z-index: 1;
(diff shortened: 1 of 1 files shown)


zuzannapaw (approved):
This error was necessary, maybe tiny but great fix :) 

I spotted one thing , I think that error message is too close to the email input. Please let me know what do you think about it 💭 

image

Anyway it works correctly 🚀

bdeluga (commented):
I remember, it generated some kind of error that's why it got removed... there was thread for it. 😅 

Edit: found it
image

wyola:
> This error was necessary, maybe tiny but great fix :)
> 
> I spotted one thing , I think that error message is too close to the email input. Please let me know what do you think about it 💭
> 
> image
> 
> Anyway it works correctly 🚀

Best I can do is ⬇️ 😅 like I wrote on slack, I think the forms' spacings etc needs revisiting in general

image

wyola:
> I remember, it generated some kind of error that's why it got removed... there was thread for it. 😅

@bdeluga do you happen to remember the thread? or what kind of error was that? i don't see anything suspicious happening after this change 👀

zuzannapaw (commented):
@wyola I agree with you. And about error message, it looks a bit overwhelming because of it's height right now, but I think that's better anyways, since it does no longer overlap the input 👀

bdeluga:
> > I remember, it generated some kind of error that's why it got removed... there was thread for it. 😅
> 
> @bdeluga do you happen to remember the thread? or what kind of error was that? i don't see anything suspicious happening after this change 👀

Log-out -> go to reports -> log-in ->go to reports

AzulloCyll:
> Log-out -> go to reports -> log-in ->go to reports

Sadly there is a problem with login again. ⬆️ log-in loop hell 😢

wyola:
> > Log-out -> go to reports -> log-in ->go to reports
> 
> Sadly there is a problem with login again. ⬆️ log-in loop hell 😢

@AzulloCyll I discussed this with @bdeluga and it should not happen anymore when:
- I merge this fix
- you merge your sign-out that redirects to login page
- @bdeluga merges his refinement monster 🐉 where he got rid of possibility of clicking on side nav when you are not logged in

we are good to go 🚀

AzulloCyll (approved):
🚀

monika-klokosinska-intive (approved):

PR #128, opened 2023-05-29
feat: mobile view category filtering #P23-365

Alessiass:
## Before ☎️ 
When layout switches to mobile category selector disappears and user is unable to add or remove existing filters.

## After 📱
After switching to mobile, accordion apears that allows the user to modify filters.
image
Accordion is created as reusable component and is available in storybook for preview.

## To be done 🗒️ 
Currently modal for managing categories is not prepared and needs to be implemented later on.

## Update 🚀 

- The state of components now properly persist after switching from mobile to desktop
- Additional 2 column view was added for tablet viewport to optimize space usage
- ForwardRef removed @Noshi96 

image


Noshi96 (changes requested):
I like this new feature, it was definitely needed! I think a better place for the management button would be the top of the list, because what if we have 40 or more categories. In that scenario, the user has to scroll a lot to get to those settings. I would also make the management button smaller. Below is my suggestion. Other than that and some minor tweaks, it works great! 🚀 

image

Vectari (approved):
Great work! 🚀
Maybe only one suggestion to get rid of that white space on tablet. What do you think?

PawelPazderski (commented):
Great work, please take a look at issue here - I noticed that if there are some categories selected and the window size changes to the mobile view no categories are selected on the list 👇 

https://github.com/intive/patronage2023-js/assets/72406958/ec55947d-e14f-43c6-ad23-919d237f4050

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
```diff
@@ -158,6 +159,7 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
           }}
         />
       </FilterSearchWrapper>
+      <MobileCategorySearch />
```
Comment (wyola) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:162]: I would put it inside the `FilterSearchWrapper`, between the button group and the search. Right now, you put it outside the wrapper, hence Card's gap comes in between the search and the accordion, so the spacing looks not great:

image

while after the change: 

image

Another thing I suggest is to change accordion's inactive border to have the same shade of grey as the search and button group have

wyola (commented):
Great addition to the app! ⭐ Two things from me

packages/ui/Accordion/Accordion.styles.tsx
```diff
@@ -0,0 +1,91 @@
...
+  ({ children, className, ...props }: any, forwardedRef) => (
+    <AccordionHeader className="AccordionHeader">
+      <AccordionTrigger
+        className={`AccordionTrigger ${className}`}
+        {...props}
+        ref={forwardedRef}>
+        {children}
+        <Icon icon={"chevron_right"} className={"iconStyles"} />
+      </AccordionTrigger>
+    </AccordionHeader>
```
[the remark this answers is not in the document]
Alessiass: This was made from RadixUI template.

packages/ui/Accordion/Accordion.styles.tsx
```diff
@@ -0,0 +1,91 @@
+import * as Accordion from "@radix-ui/react-accordion";
```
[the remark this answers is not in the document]
Alessiass: I was thinking about it, but if some functionalities would be missing later on it would be easier to extend them with RadixUI. Also it provides a lot of cool functionalities out of the hat that we don't need to think about later on, e.g. one item open at a time. Valid question 🚀

apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx
Comment (Alessiass) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx:162]: I've moved the categories to wrapper and changed colors but I've left it on the bottom since opening it moves the table down, it would be nice if the searchbar will remain in place for the whole search/filter section.

Comment (wyola) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController.tsx]: sure, you are right

apps/web/app/(navigation)/budgets/[id]/MobileCategorySearch.tsx
```diff
@@ -0,0 +1,47 @@
+import { Accordion } from "ui";
+import styled from "styled-components";
+import { device } from "lib/media-queries";
+import { useTranslate } from "lib/hooks";
+import { CategoryFilter } from "components";
```
Comment (zuzannapaw) [inline apps/web/app/(navigation)/budgets/[id]/MobileCategorySearch.tsx:5]: I suggest to change order of imports, first external ,after that internal ✔️

packages/ui/Accordion/Accordion.styles.tsx
```diff
@@ -0,0 +1,93 @@
...
+`;
+export const AccordionHeader = styled(Accordion.Header)`
+  cursor: pointer;
+`;
+export const AccordionTrigger = styled(Accordion.Trigger)`
+  display: flex;
+  justify-content: space-between;
+  cursor: pointer;
+  width: 100%;
+  font-family: unset;
```
zuzannapaw, line 34: Do we need `font-family: unset`? 👀

packages/ui/Accordion/index.tsx
```diff
@@ -0,0 +1,33 @@
+import {
+  AccordionContent,
+  AccordionItem,
+  AccordionRoot,
+  AccordionTriggerHeader,
+  StylingDiv,
+} from "./Accordion.styles";
+import React from "react";
```
zuzannapaw, line 8: I believe we don't need `import React from "react";`

zuzannapaw (approved):
Only few small things from me. It looks and works great! 🚀

packages/ui/Accordion/Accordion.styles.tsx
Alessiass: Yes, we need this to set the font to Inter.

apps/web/app/(navigation)/budgets/[id]/MobileCategorySearch.tsx
Comment (Alessiass) [reply apps/web/app/(navigation)/budgets/[id]/MobileCategorySearch.tsx]: Done 🚀

packages/ui/Accordion/index.tsx
Alessiass: I switched the import to ReactNode.

apps/web/app/(navigation)/budgets/[id]/MobileCategorySearch.tsx
```diff
@@ -0,0 +1,47 @@
...
+  font-weight: 600;
+  font-size: 0.8em;
+  align-self: end;
+`;
+
+const MobileFilter = () => {
+  const { t, dict } = useTranslate("AsideCard");
+  return (
+    <>
+      <StyledButton onClick={() => console.log("Manage modal")}>
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/MobileCategorySearch.tsx:30]: @Alessiass please remove the console.log :D.

apps/web/components/CategoryFilter/index.tsx
```diff
@@ -18,16 +18,13 @@ export const CategoryFilter = () => {
 
   const onCheckboxChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const { name, checked } = e.currentTarget;
-
+    console.log(categoryEntries);
```
monika-klokosinska-intive, line 21: @Alessiass please remember to remove the console.log :D.

packages/ui/Accordion/Accordion.styles.tsx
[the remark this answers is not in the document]
Noshi96: I have tried this without the ref and additional props, and it seems to work. What problems have you noticed with not using React.forwardRef?
```tsx
export const AccordionTriggerHeader = ({
  children,
  className,
}: React.HTMLProps<HTMLHeadingElement>) => (
  <AccordionHeader className="AccordionHeader">
    <AccordionTrigger className={`AccordionTrigger ${className}`}>
      {children}
      <Icon icon={"chevron_right"} className={"iconStyles"} />
    </AccordionTrigger>
  </AccordionHeader>
);
```

Noshi96 (approved):
It works great! I also tested on a mobile device. Good job! 🚀

apps/web/app/(navigation)/budgets/[id]/MobileCategorySearch.tsx
```diff
@@ -0,0 +1,47 @@
...
+  font-size: 0.8em;
+  align-self: end;
+`;
+
+const MobileFilter = () => {
+  const { t, dict } = useTranslate("AsideCard");
+  return (
+    <>
+      <StyledButton onClick={() => {}}>
+        ...{t(dict.categories.settings)}
```
Comment (wlechowicz) [inline apps/web/app/(navigation)/budgets/[id]/MobileCategorySearch.tsx:31]: I don't like the `...` at the beginning. It looks odd since it's followed by a capital letter. Maybe it would make sense at the end, like "Manage...", but I don't think it's necessary at all.

wlechowicz (approved):
LGTM overall, one gripe inline

wlechowicz:
One more complaint, actually. IMO "Manage" is too close to the chevron and users on mobile will fat finger the wrong thing.

obraz

bdeluga (approved):
Nice :rocket:

wyola (approved):

PR #131, opened 2023-06-01
fix: budgets list duplication #P23-383

zuzannapaw:
<h2> Problem solved 🎉  </h2>

- After adding to/removing budgets from favorites, there are no more duplicated budgets in NavList
- Duplicates are being filtered from the array

<h2> What changed 💭  </h2>

- Added superfetch for getting budget's list for proper error handling


bdeluga (approved):
Nice work  🚀

Alessiass (approved):
Looks good. Just food for thought. Whenever there is a need to store unique values we can use Sets: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

apps/web/components/SideNavigationBar/index.tsx
```diff
@@ -96,27 +97,40 @@ export default function SideNav() {
...
+
+  const removeDuplicates = (fetchedBudgets: BudgetType[]) => {
+    const uniqueBudgets = fetchedBudgets.filter(
+      (budget, idx) =>
+        fetchedBudgets
+          .map((object) => object.id.value)
+          .indexOf(budget.id.value) === idx
+    );
+    return uniqueBudgets;
+  };
```
AzulloCyll, line 127: image

You can simplify it

monika-klokosinska-intive: @Alessiass @AzulloCyll @zuzannapaw  `Set` will compare the object references, not actual id keys of the objects, so it's not a good fit here. We would have to map the array to only ids first and then use `Set`.

```
const removeDuplicates = (fetchedBudgets: BudgetType[]) => {
    return fetchedBudgets.filter(
      (budget, idx) =>
        fetchedBudgets
          .map((object) => object.id.value)
          .indexOf(budget.id.value) === idx
    );
  };
  ```
is another way to simplify it.

Comment (AzulloCyll) [reply apps/web/components/SideNavigationBar/index.tsx:127]: @monika-klokosinska-intive thank you for correcting us.

So `new Set` works on arrays of primitive data types such as number, string, booleans or nulls.
Never on array of arrays or array of objects :)

@zuzannapaw
You can resolve that conversation 😄

**Review by wyola — APPROVED** [review]
LGTM ⭐

File: apps/web/components/SideNavigationBar/index.tsx (hunk 1 above)
Comment (monika-klokosinska-intive) [reply apps/web/components/SideNavigationBar/index.tsx:127]: @AzulloCyll it will work, but by comparing the references (= addresses where the object is stored in the memory) of the objects. You could have two objects that are exactly the same, but have different references - `Set` will consider them to be two completely different instances and keep both of them. 
Usually with objects you want to compare the actual keys and values, not references, so `Set` will be useless for such  usecases. Same deal with arrays of arrays and arrays of other complex types.

But as you mentioned, primitives work as you would expect them to work, so `Set` is excellent for primitives.

@zuzannapaw please simplify the function by removing `uniqueBudgets` from inside `removeDuplicates`.

Comment (zuzannapaw) [reply apps/web/components/SideNavigationBar/index.tsx:126]: @monika-klokosinska-intive I have applied your suggestion, thank you!

**Review by Noshi96 — APPROVED** [review]
LGTM 🎸

**Review by monika-klokosinska-intive — APPROVED** [review]
@zuzannapaw Looks good! Feel free to merge :)

File: apps/web/components/SideNavigationBar/index.tsx [hunk 2]
```diff
@@ -96,27 +97,39 @@ export default function SideNav() {
...
+            isFav={isFavourite}
+            budgetId={id.value}
+            activeHref={`/budgets/${id.value}`}
+          />
+        </>
+      ),
+      href: `/budgets/${id.value}`,
+      id: id.value,
+      ref: lastBudgetRef,
+    }));
```
wlechowicz, line 116: This is a bug that's been sitting here for a while, might as well fix it now. This adds last budget ref to every single budget on the list, instead of the last one.
```suggestion
    return budgets.map(({ icon, name, id, isFavourite }, index, list) => ({
      ComponentToRender: (
        <>
          <IconStyled icon={validate(icon) ? icon : "help"} iconSize={24} />
          <SpanStyled>{name}</SpanStyled>
          <Favourite
            isFav={isFavourite}
            budgetId={id.value}
            activeHref={`/budgets/${id.value}`}
          />
        </>
      ),
      href: `/budgets/${id.value}`,
      id: id.value,
      ref: index === list.length - 1 ? lastBudgetRef : null,
    }));
```

apps/web/components/SideNavigationBar/index.tsx
```diff
@@ -96,27 +97,39 @@ export default function SideNav() {
...
+  };
+
+  const removeDuplicates = (fetchedBudgets: BudgetType[]) => {
+    return fetchedBudgets.filter(
+      (budget, idx) =>
+        fetchedBudgets
+          .map((object) => object.id.value)
+          .indexOf(budget.id.value) === idx
+    );
+  };
```
wlechowicz, line 126: This complexity of this is like O(n^2) which is bad. Here's a whole thread comparing different solutions:
https://stackoverflow.com/a/58429784/11159848
Solutions with spreading maps and sets look fancy, but don't remove any of the complexity, just hide most of it under newer operators.
One of the simplest solutions is a dictionary-based one, where you add the `id`s of the object to a dictionary as you go, and you skip ones that you already added, like this:
https://stackoverflow.com/a/18923480/11159848

wlechowicz (changes requested):

PR #133, opened 2023-06-02
feat: init personal card #P23-392

mpiwonski:
Task: [personal card task](https://tracker.intive.com/jira/browse/PATRO23-392)

**What has been done?**

- Implemented personal card component built with radix-ui components - hover card and popover (hover card >= 768px; popover < 768px - useMediaQuery from react-response)
- Added email property to session user

**How does it work?**

- Once a user opens the details of a budget, they may see other users having an access to that budget (peopleInBudget) and once they hover on their avatars, a card will pop up with information regarding their first and last name as well as their email, also a bigger avatar (desktop)
- Once a user hovers on their profile image in the navbar, a card will also pop up (similarly as with budgets) (desktop)
- When it comes to mobile users, they should tap/click on avatars instead of hovering on them
- When there's more than 4 users in a budget and when we hover on grey circle indicating the number of users that have not been displayed, we should be able to see the list of those users and then click on any position in the list and see pop-up (similar as with the ones already displayed)


packages/ui/PersonalCard/PersonalCard.styles.tsx
```diff
@@ -0,0 +1,65 @@
...
+  align-items: center;
+  background-color: ${({ theme }) => theme.personalCard.background};
+  border: 1px solid ${({ theme }) => theme.card.border};
+  box-shadow: 4px 4px 8px rgba(0, 0, 0, 0.5);
+  border-radius: 16px;
+  overflow: hidden;
+
+  padding: 0.8em 2em;
+  gap: 10px;
+  z-index: 100;
```
monika-klokosinska-intive, line 20: @mpiwonski won't anything smaller than 100 work? I know this requires extra testing with other z-indexed elements on the page, but the less you escalate the number, the fewer problems we will have in the future maintaining the code base.

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -48,6 +48,8 @@ export const MainMenu = () => {
     },
   ];
 
+  console.log(data);
```
monika-klokosinska-intive, line 51: @mpiwonski out with the console.log before merging!

packages/ui/PersonalCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  AvatarStyled,
+} from "./PersonalCard.styles";
+
+interface PersonalCardProps {
+  triggerComponent: any;
+  user: any;
+}
+
+export const PersonalCard = ({ triggerComponent, user }: PersonalCardProps) => {
+  const isTablet = useMediaQuery({ query: "(min-width: 768px)" });
```
monika-klokosinska-intive, line 21: @monika-klokosinska-intive this is a not a great idea! Better to get rid of the query check here, render both elements in the fragment and use css to decide which one will be visible on given resolution. 
Solutions like `useMediaQuery` tend to lead to... doubtful code, at least in my experience. Better to fallback on the css functionality.

aniasgit:
Merge current main, this branch is out of date.

wyola (commented):
Nice addition!

Two things from me:

1. I think that people with unreasonably long names 😢 should be taken somehow under the consideration, especially on mobile screens

image

2. There are different shadow styles on desktop and on mobile

packages/ui/PersonalCard/PersonalCard.styles.tsx
```diff
@@ -0,0 +1,65 @@
...
+`;
```
wyola, line 21: You have a lot of repetitions between `HoverCardContent` and `PopoverContent`, while you could easily simplify it like below. I smuggled there also `flex-direction: column` for the popover, since I think than on mobile the avatar should be on top and rest of the info below it (helps to solve the long name situation that I mentioned before 😅)

You could change padding to be in `px` in both cases, right now it is once `px` and once `em`.

```
/* Content styles */

const CommonContentStyled = css`
  display: flex;
  justify-content: center;
  align-items: center;
  background-color: ${({ theme }) => theme.personalCard.background};
  border: 1px solid ${({ theme }) => theme.card.border};
  box-shadow: 4px 4px 8px rgba(0, 0, 0, 0.5);
  border-radius: 16px;
  overflow: hidden;
  gap: 20px;
  z-index: 100;
`;

export const HoverCardContent = styled(HoverCard.Content)`
  ${CommonContentStyled}
  padding: 0.8em 2em;
`;

export const PopoverContent = styled(Popover.Content)`
  ${CommonContentStyled}
  padding: 10px;
  flex-direction: column;
`;
```

probably some more playing with styling is needed (like different gap etc), what's above is just a general idea

image

packages/ui/PersonalCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <HoverCardContent>
+              <AvatarStyled src={user.image ? user.image : user.avatar} />
+              <UserInfo>
+                <UserName>
+                  {user.name ? user.name : user.firstName + " " + user.lastName}
+                </UserName>
+                <UserEmail>
+                  {user.email
+                    ? user.email
+                    : user.firstName + user.lastName + `@mail.com`}
```
wyola, line 37: I know you are waiting for email to be added by backend, but maybe it is better to have here something that is clearly a placeholder, so that it is not forgotten if the BE change does not happen ;-)

packages/ui/Tooltip/index.tsx
```diff
@@ -21,6 +21,7 @@ const StyledTooltipContent = styled(RadixTooltip.Content)`
 `;
 
 export const Tooltip = ({ children, text, position }: TooltipProps) => {
+  // Instead of {text}, there might be a Personal Card component (it will receive user data ofc) taking {text} as a trigger, and once user hovers on it, hover card will be displayed?
```
wyola, line 24: I think that there is no point in mixing those two components :) tooltip is something different, if it ends up not used anywhere in the project then we can just remove it (although I think that Paweł S. end up using it in import/export)

wyola (commented):
Few more things from me

packages/ui/PersonalCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  UserInfo,
+  UserName,
+  UserEmail,
+  AvatarStyled,
+} from "./PersonalCard.styles";
+
+interface PersonalCardProps {
+  triggerComponent: any;
+  user: any;
+}
```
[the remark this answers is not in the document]
monika-klokosinska-intive: @wyola @mpiwonski firstly, this is a great catch, there is no need for an `any` here, thanks Ola!
I am on the fence where this component should go, but if it needs business logic - related stuff like `BudgetUser`, `web` might be a better place. A true atomic component would accept props like "image" "title" and "text", not a user object that is then further processed. 
Moving this to `web` is fastest and simplest solution we have.

packages/ui/PersonalCard/PersonalCard.styles.tsx
```diff
@@ -0,0 +1,65 @@
...
+  font-size: 24px;
+`;
+
+export const UserEmail = styled.span`
+  font-size: 12px;
+`;
+
+export const AvatarStyled = styled(Avatar)`
+  height: 96px;
+  width: 96px;
```
zuzannapaw, line 43: That's only my point of view, but I'm not sure about avatar's size. Comparing to other elements on page (specially the main ones, that are more exposed), it seems a bit too overwhelming for me.  

image

This is how it would look like with a bit smaller dimensions, like 70px. 

image

I think it look more neat, but please tell me what do you think about it! 👀

packages/ui/PersonalCard/PersonalCard.styles.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const PopoverContent = styled(Popover.Content)`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-color: ${({ theme }) => theme.personalCard.background};
+  border: 1px solid ${({ theme }) => theme.card.border};
+  box-shadow: 0 2px 6px rgba(32, 41, 50, 0.1);
+  border-radius: 16px;
+  overflow: hidden;
+
```
zuzannapaw, line 57: look out for gaps 👀

packages/ui/PersonalCard/index.tsx
zuzannapaw, line 37: Instead of adding these strings together, you could simply use template string. It would look like:

```javascript
 {user.name ? user.name : `${user.firstName } ${user.lastName}`}
``` 
```javascript
{user.email ? user.email : `${user.firstName} ${user.lastName} ${@mail.com}`}
```

packages/ui/PersonalCard/PersonalCard.styles.tsx
zuzannapaw: I think there will be more problems with z-index in this case:
image

packages/ui/PersonalCard/PersonalCard.styles.tsx
```diff
@@ -0,0 +1,65 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-color: ${({ theme }) => theme.personalCard.background};
+  border: 1px solid ${({ theme }) => theme.card.border};
+  box-shadow: 4px 4px 8px rgba(0, 0, 0, 0.5);
+  border-radius: 16px;
+  overflow: hidden;
+
+  padding: 0.8em 2em;
```
zuzannapaw, line 18: I'm not sure if it's better for paddings to be set in `px`, it gives more control when being on different devices

zuzannapaw (commented):
Great addition to the app ☺️ ! Few thing from me.
I also spotted one thing - was it intended to add yellow border on mobile view when the element on focus? 

image

Well done for now! 💪

packages/ui/PersonalCard/PersonalCard.styles.tsx
mpiwonski: Thanks @wyola for your insights. These are really thoughtful :)

packages/ui/PersonalCard/index.tsx
mpiwonski: We discussed that during today's call, however we had not come up with a clear answer how to handle that. As for now, it may remain as it is :)

packages/ui/PersonalCard/index.tsx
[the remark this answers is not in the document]
mpiwonski: Thanks for the points you have raised. I do agree with you, @monika-klokosinska-intive, that it should receive props and there should not be any special operations to decide what to render inside, although keep in mind there's data discrepancy related to the same data (logically) like image and avatar; name -> firstName and lastName...

mpiwonski:
> Great addition to the app ☺️ ! Few things from me.
> 
> I also spotted one thing - was it intended to add yellow border on mobile view when the element on focus?
> 
> image
> 
> Well done for now! 💪

I guess it's a built-in feature provided by radix for this component :P

packages/ui/PersonalCard/PersonalCard.styles.tsx
mpiwonski: You may be actually right :) I've changed it to 80px :)

packages/ui/PersonalCard/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  image: string;
+}
+
+const getPersonalCardContent = (name: string, email: string, image: string) => {
+  return (
+    <>
+      <AvatarStyled src={image} />
+      <UserInfo>
+        <UserName>{name}</UserName>
+        <UserEmail>{email ? email : `${name}@mail.com`}</UserEmail>
```
aniasgit, line 30: If you change email property for optional, you can only display email if it exists

```suggestion
        {email && <UserEmail>{email}</UserEmail>}
```

monika-klokosinska-intive: @mpiwonski this is what I would go for as well.

mpiwonski: Thanks, I have incorporated that change :)

zuzannapaw (approved):
Works great! Nice job! 🚀

Alessiass (approved):
I found small bug -> When you turn on popover on mobile and screen size changes it remains in top left corner. Besides of that it work's fine.
image

packages/ui/package.json
```diff
@@ -22,14 +22,16 @@
...
+    "@radix-ui/react-hover-card": "^1.0.6",
+    "@radix-ui/react-popover": "^1.0.6",
     "@radix-ui/react-select": "^1.2.1",
     "@radix-ui/react-tabs": "^1.0.3",
     "@radix-ui/react-toast": "^1.1.3",
     "@radix-ui/react-tooltip": "1.0.5",
     "material-symbols": "^0.5.5",
     "react-datepicker": "^4.11.0",
     "react-loading-skeleton": "^3.3.1",
+    "react-responsive": "^9.0.2",
```
monika-klokosinska-intive, line 34: @mpiwonski you can remove it from package.json if it's not needed anymore :)

monika-klokosinska-intive (approved):
@mpiwonski I am giving a tentative approval here to speed things up, but please:
- resolve conflicts with main branch
- remove the unused library  
- @zuzannapaw makes a good point about using `px` , please change the `em` to `px`

wyola (changes requested):
Two things from me:

- something changed here and avatars are overflowing the container on mobile. It is not like it on main, so it is matter of this branch (this is why I am requesting changes)

image

image

- when you hover on the grey circle where more users are hiding, and then you display personal card of the first user, you can move to the second one. Idk if there is a quick fix for this the way it is designed now. Probably both Hover Card and Popover have some prop like `side` where you can pass on which side it should display. In this case I would display it on the left, but then on mobile it should be on the right... idk


https://github.com/intive/patronage2023-js/assets/80450315/2c03f874-5c75-40c2-a46e-c11ad5545960

wyola:
> I found small bug -> When you turn on popover on mobile and screen size changes it remains in top left corner. Besides of that it work's fine. image

@mpiwonski I've had similar problem with `DropdownMenu` https://github.com/intive/patronage2023-js/pull/119#pullrequestreview-1449199360

check this fix - https://github.com/intive/patronage2023-js/pull/119/commits/617b71b9d43199bba26ee1b27a0a747462601bee

apps/web/app/(navigation)/MainMenu.tsx
```diff
@@ -59,8 +59,15 @@ export const MainMenu = () => {
           onValueChange={menuHandler}
           onOpenChange={() => setIsIconDown(!isIconDown)}>
           <SelectTriggerStyled>
-            <Select.Value>
-              <AvatarStyled src={data.user.image} outlined />
+            <Select.Value asChild>
+              <PersonalCard
+                triggerComponent={
+                  <AvatarStyled src={data.user.image} outlined />
```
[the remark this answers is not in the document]
mpiwonski: Thanks for your suggestions. I have included them. Please mark this conversation as resolved.

packages/ui/PersonalCard/PersonalCard.styles.tsx
Noshi96, line 1: When we have a long username and email address:
image

https://github.com/intive/patronage2023-js/blob/eb02a9e117d38272c4af80cb74bc88b43935e720/packages/ui/Tooltip/index.tsx#L13-L22
change to `z-index: 5;`

packages/ui/PersonalCard/PersonalCard.styles.tsx
```diff
@@ -0,0 +1,97 @@
...
+`;
+
+export const PopoverContent = styled(Popover.Content)`
+  ${CommonContentStyled}
+  flex-direction: column;
+
+  ${device.tablet} {
+    display: none;
+  }
+`;
```
Noshi96, line 87:
```suggestion
export const PopoverContent = styled(Popover.Content)`
  ${CommonContentStyled}
  flex-direction: column;
  max-width: 100vw;

  ${device.tablet} {
    display: none;
  }
`;
```
My suggestion:
image

Old one:
image

https://github.com/intive/patronage2023-js/blob/2056f416fa542f8d4f16ea4c88701ed951263da5/apps/web/app/(navigation)/budgets/%5Bid%5D/BudgetBasinInformation/PeopleInBudget.tsx#L72

You can change to "top"

https://github.com/intive/patronage2023-js/blob/2056f416fa542f8d4f16ea4c88701ed951263da5/apps/web/components/HiddenUsers/index.tsx#L58

change to "bottom"

Noshi96 (approved):
I missed the function of viewing people assigned to a budget on a smartphone, and now thanks to your popover this is possible. I like how it works and looks. 🚀 
I have a few minor suggestions, especially in edge cases such as long usernames and emails

wyola (approved):

aniasgit (approved):
Looks ands works good 👍

PR #135, opened 2023-06-02
fix: no redirect on sign in #P23-387

wyola:
## Fix for no redirect on sign-in
task: https://tracker.intive.com/jira/browse/PATRO23-387

### What changed 🔨 
- user should be now redirected correctly after pressing the "sign in" button

### Why it stopped working 🤔
It looks like on production build after login, Next Router needs to be refreshed to access routes to which user previously had no access (`/`) (route `/` was recently changed to be protected).

To reproduce this error locally, you have to run production build locally:
- go to `apps/web` folder in terminal and 
- run commands `npm run build` then `npm run start`

### Other fixes 🔧 
- if user would logout while being on one of the "settings" subpages, then they would end up not on login page but on "settings" page, but without the side nav bar
- login form was sending request twice, since the button is of `type="submit"` and had also `onClick={submit}`

### DIFF
--- a/apps/web/app/(regflow)/sign-in/page.tsx
+++ b/apps/web/app/(regflow)/sign-in/page.tsx
@@ -64,13 +64,14 @@ export default function SignInPage() {
 
   return (
     <Form
-      onSubmit={async (values) => {
+      onSubmit={(values) => {
         signIn("credentials", {
           email: values.email,
           password: values.password,
           redirect: false,
         }).then((res) => {
-          if (res!.ok) {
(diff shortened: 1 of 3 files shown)


wlechowicz (approved):

AzulloCyll (approved):
Work done for at least 120%. 🚀

bdeluga (approved):
Nice 🚀

Noshi96 (approved):
🎸

Vectari (approved):
🚀

PR #136, opened 2023-06-03
fix: now should properly calculate range

bdeluga:
Fixed logic for range calculation, the bug was happening when we were looking at older budgets 🚀 
Edit: also since we are able to create transaction in the future i changed how how week and month are calculated


monika-klokosinska-intive (approved):

AzulloCyll (approved):
🚀

PR #138, opened 2023-06-03
feat: handle validation codes from api #P23-377

Vectari:
# Hello! 👋
BE is not ready so I open this PR as draft. [P23-377](https://tracker.intive.com/jira/browse/PATRO23-377).
Error message is hardcoded but support all languages 🥳
## About
Some error user will never see because frontend validation is awesome but other errors show in toasts. Error with code 10.X show in console because this is programming class error.
## How to use it?
1. Now error message is hardcoded but you can change `44` line of useSuperFetch.ts `const hardcodedError = "10.10";` to see toast with another error.
2. Break something somewhere where superFetch is used to see error message (for example TransactionTableController 🙃)

### Have a nice day!

### DIFF
--- a/apps/web/lib/dictionary.ts
+++ b/apps/web/lib/dictionary.ts
@@ -1073,6 +1073,194 @@ const dictionary = {
       fr: "Remboursement",
     },
   },
+  Errors: {
+    "1.1": {
+      //FE validation
+      en: "Budget name is too long.",
+      pl: "Nazwa budżetu jest zbyt długa.",
+      fr: "Le nom du budget est trop long.",
+    },
+    "1.2": {
+      //FE validation
+      en: "Name cannot be empty.",
+      pl: "Nazwa nie może być pusta.",
+      fr: "Le nom ne peut pas être vide.",
+    },
+    "1.3": {
+      //FE validation
+      en: "Name is too long.",
+      pl: "Nazwa jest zbyt długa.",
+      fr: "Le nom est trop long.",
+    },
+    "1.4": {
+      //TOAST
+      en: "Budget with given name already exists.",
+      pl: "Budżet o podanej nazwie już istnieje.",
+      fr: "Un budget portant le nom donné existe déjà.",
+    },
+    "1.5": {
+      //FE validation
+      en: "Start date cannot be empty.",
+      pl: "Data rozpoczęcia nie może być pusta.",
+      fr: "La date de début ne peut pas être vide.",
+    },
+    "1.6": {
+      //FE validation
+      en: "End date cannot be empty.",
+      pl: "Data zakończenia nie może być pusta.",
+      fr: "La date de fin ne peut pas être vide.",
+    },
+    "1.7": {
+      //FE validation
+      en: "Start date must be earlier than end date.",
+      pl: "Data rozpoczęcia musi być wcześniejsza niż data zakończenia.",
+      fr: "La date de début doit être antérieure à la date de fin.",
+    },
+    "1.8": {
+      //FE validation
+      en: "Limit cannot be empty.",
+      pl: "Limit nie może być pusty.",
+      fr: "La limite ne peut pas être vide.",
+    },
+    "1.9": {
+      //FE validation
+      en: "Limit must be greater than 0.",
+      pl: "Limit musi być większy niż 0.",
+      fr: "La limite doit être supérieure à 0.",
+    },
+    "1.10": {
+      //Not working with hardcoded currency
+      en: "Currency is not supported.",
+      pl: "Nie obsługiwana waluta.",
+      fr: "La monnaie n'est pas supporté.",
+    },
+    "1.11": {
+      //TOAST
+      en: "Budget not exists.",
+      pl: "Budżet nie istnieje.",
+      fr: "Le budget n'existe pas.",
+    },
+    "2.1": {
+      //FE validation
+      en: "Transaction type cannot be empty.",
+      pl: "Typ transakcji nie może być pusty.",
+      fr: "Le type de transaction ne peut pas être vide.",
+    },
+    "2.2": {
+      //TOAST
+      en: "Invalid budget transaction type.",
+      pl: "Nieprawidłowy typ transakcji budżetu.",
+      fr: "Type de transaction invalid",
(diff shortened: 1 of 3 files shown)


apps/web/lib/dictionary.ts
```diff
@@ -1073,6 +1073,215 @@ const dictionary = {
...
+      fr: "Le nom de la transaction est trop long.",
+    },
+    "2.5": {
+      //FE validation
+      en: "Value cannot be empty.",
+      pl: "Wartość nie może być pusta.",
+      fr: "La valeur ne peut pas être vide.",
+    },
+    "2.6": {
+      //FE validation - FE add positive for income or negative for expense
```
monika-klokosinska-intive, line 1179: @Vectari `FE adds` ;)

apps/web/lib/dictionary.ts
```diff
@@ -1073,6 +1073,215 @@ const dictionary = {
...
+      en: "Value must be positive for income or negative for expense.",
+      pl: "Wartość musi być dodatnia dla przychodu lub ujemna dla wydatku.",
```
aniasgit, line 1181: In our app we use "wpływ" not "przychód" in polish.

apps/web/lib/hooks/useSuperfetch.ts
```diff
@@ -35,9 +40,31 @@ export default function useSuperfetch() {
...
           const data = await res.json().catch(() => ({}));
           return { ...data, httpStatus: res.status };
+        } else if (res.status === 400) {
+          const hardcodedError = "10.10";
+          showToast({
+            variant: "error",
+            message:
+              t(dict.title) + hardcodedError + " - " + t(dict[hardcodedError]),
+          });
+        } else if (res.status !== 200 || 400 || 401) {
```
aniasgit, line 50: There are places where the ok status is 201 (editBudget, createNewBudget, createNewTransaction)

wyola, line 50: I am a bit confused about what this line does 👀 you are not really comparing anything here, since this will always be truthy (`400` is always truthy). Maybe you can simplify it like below. You already check for 200 and 201 in the first if `(res.ok)`, then for 400 in the else if condition. 

```suggestion
        } else if (res.status > 401) {
```

apps/web/lib/dictionary.ts
Vectari: Fixed :)

apps/web/lib/dictionary.ts
Vectari: Fixed :)

apps/web/lib/hooks/useSuperfetch.ts
Vectari: I changed it like you suggested but I have to test it more when BE will be ready.

apps/web/lib/dictionary.ts
```diff
@@ -1073,6 +1073,215 @@ const dictionary = {
...
+    "1.9": {
+      //FE validation
+      en: "Limit must be greater than 0.",
+      pl: "Limit musi być większy niż 0.",
+      fr: "La limite doit être supérieure à 0.",
+    },
+    "1.10": {
+      //Not working with hardcoded currency
+      en: "Currency is not supported.",
+      pl: "Nie obsługiwana waluta.",
```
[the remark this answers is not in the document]
Vectari: My bad... 🤦‍♂️

apps/web/lib/hooks/useSuperfetch.ts
Vectari: https://github.com/intive/patronage2023-js/pull/138#discussion_r1218262753

apps/web/lib/hooks/useSuperfetch.ts
```diff
@@ -35,9 +39,31 @@ export default function useSuperfetch() {
...
+            variant: "error",
+            message: t(dict["defaultError"]),
+          });
         }
-        throw new SuperError("An error has occurred.", res.status);
+        // CODE BELOW IS DRAFT AND WAIT FOR BE
+        // } else if ((res.status === 400 && ErrorCode === "10.1") || "10.2") {
+        //   console.log(
+        //     t(dict.title) + " " + `${[ErroCode]}` + ": " + t(dict[ErrorCode])
+        //   );
```
zuzannapaw, line 59: For the future, when BE will be ready, instead of adding strings together, you could simply use template string for all of these values 

```javascript
`${t(dict.title)} ${[ErroCode]} : ${t(dict[ErrorCode])}`
```

apps/web/lib/hooks/useSuperfetch.ts
```diff
@@ -35,9 +39,31 @@ export default function useSuperfetch() {
         if (res.ok) {
           const data = await res.json().catch(() => ({}));
           return { ...data, httpStatus: res.status };
+        } else if (res.status === 400) {
+          const hardcodedError = "10.10";
+          showToast({
+            variant: "error",
+            message:
+              t(dict.title) + hardcodedError + " - " + t(dict[hardcodedError]),
```
zuzannapaw, line 47: Same here with template string:
```javascript
`${t(dict.title)} ${hardcodedError}  -  ${t(dict[hardcodedError])}`
```

zuzannapaw (commented):
Nice work! ✨ Here are few suggestions from me. Good job 💪

Alessiass (approved):
LGTM 🚀

apps/web/lib/dictionary.ts
```diff
@@ -1104,6 +1104,215 @@ const dictionary = {
       fr: "Remboursement",
     },
   },
+  Errors: {
```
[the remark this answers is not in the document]
monika-klokosinska-intive: @Vectari @zuzannapaw not a bad idea, it would definitely make working with errors easier in the future (because we can assume as the app would grow, so would the amount of custom errors codes). However since we are wrapping up, I don't believe this is a blocker.

monika-klokosinska-intive (approved):

Noshi96 (approved):
LGTM 🎸

wyola (approved):
⭐

PR #141, opened 2023-06-05
feat: reports page #P23-371

PawelPazderski:
## Reports page 🎆 
ticket: https://tracker.intive.com/jira/browse/PATRO23-371

Reports page displaying accumulated information from all budgets. 
Please log in as 
_reporter@gmail.com_ 
and use the password same as Mr Jarzyna has - 
I created few budgets on this account to see data for the available periods the user can choose from:
- two budgets in USD
- one budget in PLN - no transactions when choosing 7 days to display the information accordingly
- no budgets in EUR - display info that there are no budgets in this currency

The page by default sends a request for budgets n USD for 12 months, also the default 12 months go with the request after currency change.
- charts provide full language support (the months translation is delivered by dayjs) - need some french translations though @zuzannapaw - when you find a moment ☺️ 
- chart component displays a chart depending on user's choice - could be line chart or bar chart
- I set the charts to max-height 50vh - IMO it's nice to see it on the screen together with the options and information, it still behaves well on all width range and is fully responsive
- the way things are displayed - tooltip, custom line, dates, amounts and information is accepted by Client
- line chart has a custom line that goes together with a tooltip (there is a little problem here, didn't find the solution so far or maybe it is not that bad and deadly - typescript has a problem with the property that exists and what is more - determines proper behaviour that the line shows only when tooltip is active - it all works fine - but TS says here:
`Property '_active' does not exist on type 'TooltipModel<keyof ChartTypeRegistry>'. Did you mean 'active'?`
maybe some TS specialists can give me a hint here I tried really long to walk this over...

image

The group button to choose period is a new component ButtonGroupSimple, I tried to reuse the ButtonGroup but:
1. Had a problem using both at same time (the checked option did't stay where it should) - didn't want to overwrite the logic as it is used elsewhere also because of 👇 
2. Many differences in appearance so decided to write new component instead of adding lot's of conditions to the ButtonGroup - which ends up with two components cooperating as expected (keeping the state, checked behaviour etc)

~~Not quite good yet:~~
~~BE gives back null for trendValue in 12 month option~~
Edit: my bad there was data only within the asked period so the account was 0 so there was no way to count trend value. 

Hope you like it, had tons of issues to solve on the way starting with the fact that the currency option was not taken into account at the beginning, and in this case is crucial for the correctness of the data.. 

Possible to do:
- add suspense would be nice, now there is only a spinner


apps/web/app/(navigation)/reports/page.tsx
```diff
@@ -1,5 +1,311 @@
...
+    tag: "EURO",
+    locale: "de-DE",
+  },
+};
 
 export default function ReportsPage() {
-  return <ReportsTitle />;
+  const { t, dict } = useTranslate("ReportsPage");
+  const { title, aside, balance, info, currency, currencyNames } = dict;
+  const [chart, setChart] = useState("line");
```
Noshi96, line 55: We can add type to this 
```tsx
type ChartTyp = "line" | "bar";
```

apps/web/app/(navigation)/reports/page.tsx
```diff
@@ -1,5 +1,311 @@
...
+        <PageWrapper>
+          <TopWrapper>
+            <TitleStyled>{t(title)}</TitleStyled>
+            <ButtonGroupSimple
+              options={[
+                {
+                  component: "7d",
+                  id: "7days",
+                  checked: timeRange === "7days",
+                },
```
Noshi96, line 189: If you have considered using `TimeRangeType`, there are still a few changes here and inside `ButtonGroupSimple` to make :D
```tsx
{
  component: "7d",
  id: "7days",
  tRange: { period: 7, unit: "days" },
};
```

packages/ui/ButtonGroupSimple/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  component: ReactNode;
+  id: string;
+  checked: boolean;
+};
+
+type GroupProps = {
+  options: InputProps[];
+  selectedOption: string;
+  onOptionSelect: (id: string) => void;
+} & React.HTMLProps<HTMLDivElement>;
```
Noshi96, line 17: If you have chosen `TimeRangeType`
```tsx
type InputProps = {
  component: ReactNode;
  id: string;
  tRange: TimeRangeType;
};

type GroupProps = {
  options: InputProps[];
  selectedOption: TimeRangeType;
  onOptionSelect: (timeRange: TimeRangeType) => void;
```

packages/ui/ButtonGroupSimple/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  label {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    cursor: pointer;
+    padding: 4px 2px;
+    position: relative;
+    text-align: center;
+    border-bottom: 2px solid transparent;
+    color: #7e7e7e;
```
Noshi96, line 36: You can add color to the theme

packages/ui/ButtonGroupSimple/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+    z-index: 1;
+  }
+`;
+
+export const ButtonGroupSimple = ({
+  options,
+  selectedOption,
+  onOptionSelect,
+}: GroupProps) => (
+  <ButtonRadioGroupStyled>
```
Noshi96, line 71: I'm thinking of adding some arias here
```suggestion
  <ButtonRadioGroupStyled role="radiogroup" aria-orientation="vertical">
```

packages/ui/ButtonGroupSimple/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+    {options.map(({ component, id }) => (
+      <label key={id}>
+        <input
+          type="radio"
+          value={id}
+          checked={selectedOption === id}
+          onChange={() => onOptionSelect(id)}
+        />
```
Noshi96, line 79: If you have chosen `TimeRangeType`
```tsx
  options.map(({ component, id, tRange }) => (
    <label key={id}>
      <input
        type="radio"
        value={id}
        checked={`${selectedOption.period}${selectedOption.unit}` === id}
        onChange={() => onOptionSelect(tRange)}
      />
      {component}
    </label>
```

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,276 @@
...
+        borderColor: theme.reports.incomesLine,
+        backgroundColor: backgroundIncomes,
+        borderRadius: 5,
+        fill: true,
+      },
+      {
+        label: t(labelsTooltip.expenses),
+        data: expenses,
+        borderColor: theme.reports.expencesLine,
+        backgroundColor: backgroundExpences,
```
Noshi96, line 115: typo

Noshi96 (commented):
I can see a lot of work and effort put into this task. The result is magnificent! Great UX. I see a few improvements that could be considered, but other than that, great job!  ✨ ✨ ✨ 

P.S. I think the chart type buttons are too big
P.S2. In the Jira task there is a point 7 about the total balance reflected in the selected period, does this work?
P.S3. There is an authorization problem when refreshing the page in reports
image

PawelPazderski:
> I can see a lot of work and effort put into this task. The result is magnificent! Great UX. I see a few improvements that could be considered, but other than that, great job! ✨ ✨ ✨
> 
> P.S. I think the chart type buttons are too big P.S2. In the Jira task there is a point 7 about the total balance reflected in the selected period, does this work? P.S3. There is an authorization problem when refreshing the page in reports image

Thanks a lot, it took some work indeed. 
- you're right - the buttons are too big, I'm changing it
- the word "total" messed my mind here a bit - you're right that should print the value for the chosen period not all budgets (BE gives both of those in response)
- not sure what couses the error so far

packages/ui/ButtonGroupSimple/index.tsx
PawelPazderski: I really like your the idea to devide the timeRange in general but I think I would leave it as it is, the button group with the id only is more reusable. After all, there are only six options in the switch and few if statements.

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+  Filler,
+  BarController,
+  BarElement,
+  ChartOptions,
+  ChartTypeRegistry,
+  Tooltip,
+  TooltipItem,
+  TooltipModel,
+} from "chart.js";
+import { Line, Bar } from "react-chartjs-2";
```
zuzannapaw, line 20: I suggest to put external imports first , after that our internal imports

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+  //using colors with high transparency to create gentle gradients
+  const gradientGreen = {
+    startColor: "rgba(0, 200, 0, 0.2)",
+    endColor: "transparent",
+  };
+
+  const gradientGrey = {
+    startColor: "rgba(0, 0, 0, 0.2)",
+    endColor: "transparent",
+  };
```
zuzannapaw, line 76: Maybe a good idea would be to add these colors to the theme? 🤔

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+        chartArea: { top, bottom },
+      } = chart;
+
+      //can't find a way to type here properly, tried for very long time, but works as expected and has to be "_active" not "active" as suggested by ts
+      //had a solution but the line didn't disapear after leaving chart:(
+      if (tooltip && tooltip._active[0]) {
+        const activeTooltip = tooltip.dataPoints[0];
+
+        ctx.beginPath();
+        ctx.strokeStyle = "grey";
```
zuzannapaw, line 140: it could be moved as well to the theme 👍

apps/web/app/(navigation)/reports/ReportsPage.styled.tsx
```diff
@@ -0,0 +1,94 @@
...
+export const StyledWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  font-size: 12px;
+  line-height: 150%;
+  color: ${({ theme }) => theme.trendChart.titleLeft};
+  border: 1px solid ${({ theme }) => theme.infoTile.border};
+  border-radius: 8px;
+  padding: 1em 2em;
```
zuzannapaw, line 38: I've heard that setting paddings in other values that `px` can cause some problems on devices with different sizes. Setting values in `px` can give more safety and precise control

zuzannapaw (commented):
Great feature, superb ui ✨  Only few small suggestions from me

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+        },
+        beginAtZero: true,
+        ticks: {
+          //change chart step depending on average value
+          stepSize: (): number => {
+            const sum = data.datasets[0].data.reduce(
+              (acc, val) => acc + val,
+              0
+            );
+            const average = sum / data.datasets[0].data.length;
```
wyola, line 212: I admit I did not dig into details but I assume you checked that this can't end up in dividing by zero ;-)

apps/web/app/(navigation)/reports/ReportsChart.tsx
PawelPazderski: yep, thanks

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+            return average > 10000 ? 10000 : 1000;
```
wyola, line 213: I think that step should be calculated differently. What if there is a situation when both expenses and incoms are rather low? Then there will be a lot of white space and no grid for those values, since first line will be at 1K. You can calculate max value from the data and use it to calculate the step. 

some example with a lot of white space at the top part of the chart:
image

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+          },
+          //set custom format 1000 => 1K, 10000 => 10K etc
+          callback: (value: number) => {
+            if (value >= 1000) {
+              return value / 1000 + " K";
```
wyola, line 218: `K` should be in dictionary - `K` is in English, `tys` in Polish, something else in french probably ;-)

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+              ref={chartRef}
+              style={{ maxHeight: "50vh" }}
+              data={data}
+              options={lineOptions as ChartOptions<"line">}
+              plugins={linePlugins}
+            />
+          )}
+          {chart === "bar" && (
+            <Bar
+              style={{ maxHeight: "50vh" }}
```
wyola, line 266: you can pass `className` here and don't hardcode styles in the component but move them to the styles files

apps/web/app/(navigation)/reports/ReportsChart.tsx
PawelPazderski: ok, need to define some new transparency versions but sounds good

apps/web/app/(navigation)/reports/page.tsx
```diff
@@ -1,5 +1,317 @@
...
+            // If it's an income transaction, add the value to the incomes total
+            acc[day].incomes += value;
+          } else {
+            // If it's an expense transaction, add the value to the expenses total
+            acc[day].expenses += value;
+          }
+
+          return acc;
+        }, {});
+    }
```
wyola, line 174: you could simplify this, since those two reducers do the same thing

```
 const dateFormat =
      timeRange === "12month" ||
      timeRange === "6month" ||
      timeRange === "3month"
        ? "MMM | YY"
        : "DD-MM";
    transactions = [...statistics.incomes, ...statistics.expenses].reduce(
      (acc, transaction) => {
        const date = dayjs(transaction.datePoint);
        const grouping = date.format(dateFormat);
        const { value } = transaction;

        if (!acc[grouping]) {
          // If the month is not yet in the accumulator, initialize it with zero income and expense
          acc[groupping] = { incomes: 0, expenses: 0 };
        }

        if (statistics.incomes.includes(transaction)) {
          // If it's an income transaction, add the value to the incomes total
          acc[grouping].incomes += value;
        } else {
          // If it's an expense transaction, add the value to the expenses total
          acc[grouping].expenses += value;
        }

        return acc;
      },
      {}
    );
```

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+  return (
+    <>
+      {Object.keys(transactions).length === 0 ? (
+        <h3>{t(info.noTransaction)}</h3>
+      ) : (
+        <>
+          {chart === "line" && (
+            <Line
+              ref={chartRef}
+              style={{ maxHeight: "50vh" }}
```
wyola, line 258: you can pass `className` here and don't hardcode styles in the component but move them to the styles files

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+                case "USD":
+                  value = "$ " + context.parsed.y.toLocaleString();
+                  break;
+                case "PLN":
+                  value = context.parsed.y.toLocaleString() + " PLN";
+                  break;
+                case "EUR":
+                  value = "€ " + context.parsed.y.toLocaleString();
+                  break;
+              }
```
wyola, line 177: formatting is wrong here and should be the same as is in `CurrencyAmount` IMO, so:
- $ in front are OK
- PLN -> `CurrencyAmount` displays `zł`
- € should be at the back

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+    beforeDatasetsDraw(chart: any) {
+      const {
+        ctx,
+        tooltip,
+        chartArea: { top, bottom },
+      } = chart;
+
+      //can't find a way to type here properly, tried for very long time, but works as expected and has to be "_active" not "active" as suggested by ts
+      //had a solution but the line didn't disapear after leaving chart:(
+      if (tooltip && tooltip._active[0]) {
```
wyola, line 135: I think that the problem here is the fact that you try to use "private" property. Those are usually (to my best knowledge) indicated by the underscore at the beginning and can't be really accessed - this is why TS says that there is no such property. But I am not sure of how to fix and if what I am writing is 100% true :(

wyola (commented):
Ufff, lots of work on your side! ⭐ 

I left a few suggestions and ideas. One thing that was irritating while clicking through the charts is the fact that when I change the currency, the budget period resets. I think it should keep the last choice.

apps/web/app/(navigation)/reports/page.tsx
Noshi96: I agree with @wyola ☝️

apps/web/app/(navigation)/reports/ReportsChart.tsx
PawelPazderski: that unfortunately ruins the responsiveness of the chart tried to put them in containers/wrappers and didn't work well

Noshi96:
> * not sure what couses the error so far

It looks like the fetch is done before the access token is received, we can use our old cheat here to overcome this
```tsx
import { useSession } from "next-auth/react";
  const { data: statistics, isLoading } = useQuery({
    queryKey: ["budgetsStatistics", start, end, reportsCurrency, language],
    queryFn: async () => {
      return fetch(
        `${env.NEXT_PUBLIC_API_URL}budgets/statistics?startDate=${start}&endDate=${end}&currency=${reportsCurrency}`
      ).catch((err) => console.error(err));
    },
    enabled: !!session,
  });
```

apps/web/app/(navigation)/reports/page.tsx
PawelPazderski: they don't do exactly the same thing, in case of days we need also :
```
...

.sort((a, b) => dayjs(a.datePoint).valueOf() - dayjs(b.datePoint).valueOf())

...
```
which is necessary here

apps/web/app/(navigation)/reports/ReportsChart.tsx
PawelPazderski: I believe I can't same as with the Bar chart - loosing responsiveness

apps/web/app/(navigation)/reports/ReportsChart.tsx
PawelPazderski: yep, correcting that

apps/web/app/(navigation)/reports/page.tsx
wyola: I've seen the sort, but when I was checking it everything seemed to work good without it 🤔 will look at it later once again

apps/web/app/(navigation)/reports/ReportsChart.tsx
wyola: @PawelPazderski I meant like so:

```
import { Line } from "react-chartjs-2";

export const StyledLine = styled(Line)`
  max-height: 50vh;
`;
```

and then:

```
        {chart === "line" && (
            <StyledLine
              ref={chartRef}
              data={data}
              options={lineOptions as ChartOptions<"line">}
              plugins={linePlugins}
            />
```
it seems to work fine for me, but maybe not very worth it for two lines of styles

apps/web/app/(navigation)/reports/page.tsx
wyola: @PawelPazderski oh okay, you are right, now I see it does not work, I must have come across some very organized data before 😅 (or simply can't see properly anymore)

Nevertheless, you could still simplify and have ~20 lines less of code:

```
   const groupByMonth =
      timeRange === "12month" ||
      timeRange === "6month" ||
      timeRange === "3month";
    const dateFormat = groupByMonth ? "MMM | YY" : "DD-MM";

    transactions = [...statistics.incomes, ...statistics.expenses];

    if (!groupByMonth) {
      transactions = transactions.sort(
        (a, b) => dayjs(a.datePoint).valueOf() - dayjs(b.datePoint).valueOf()
      );
    }

    transactions = transactions.reduce((acc, transaction) => {
      const date = dayjs(transaction.datePoint);
      const grouping = date.format(dateFormat);
      const { value } = transaction;

      if (!acc[grouping]) {
        // If the month is not yet in the accumulator, initialize it with zero income and expense
        acc[grouping] = { incomes: 0, expenses: 0 };
      }

      if (statistics.incomes.includes(transaction)) {
        // If it's an income transaction, add the value to the incomes total
        acc[grouping].incomes += value;
      } else {
        // If it's an expense transaction, add the value to the expenses total
        acc[grouping].expenses += value;
      }

      return acc;
    }, {});
```

wyola (commented):
forgot about two things

apps/web/app/(navigation)/reports/page.tsx
PawelPazderski: that's looks great, thanks 🚀

apps/web/app/(navigation)/reports/ReportsChart.tsx
PawelPazderski: ok, I thought of wrappers but that's fine, works as should and keeps styling in one place, let's do so 
thanks a lot!

apps/web/app/(navigation)/reports/ReportsChart.tsx
```diff
@@ -0,0 +1,278 @@
...
+    const gradientFill = ctx!.createLinearGradient(0, 0, 0, 400);
+    gradientFill.addColorStop(0, gradient.startColor);
+    gradientFill.addColorStop(1, gradient.endColor);
+    return gradientFill;
+  };
+
+  //define background colors depending on chart
+  const backgroundIncomes =
+    chart === "line" ? createGradient(gradientGreen) : theme.reports.incomesBar;
+  const backgroundExpences =
```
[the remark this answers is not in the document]
PawelPazderski: ok corrected

apps/web/app/(navigation)/reports/ReportsChart.tsx
PawelPazderski: good point, setting K, tys, k for french

apps/web/app/(navigation)/reports/page.tsx
```diff
@@ -1,5 +1,318 @@
...
+                <StyledInfo>
+                  {t(info.noData)}
+                  {reportsCurrency}
+                </StyledInfo>
+              </StyledReportsBalanceWrapper>
+            )}
+          </StyledWrapper>
+          <ButtonWrapper>
+            <SelectWrapper>
+              <Select
```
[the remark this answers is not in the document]
PawelPazderski: ok done

apps/web/app/(navigation)/reports/ReportsChart.tsx
PawelPazderski: wrote condition just in case

apps/web/app/(navigation)/reports/ReportsChart.tsx
zuzannapaw: @PawelPazderski okay! Im sure there is a solution for that, but it can stay the same, you can ignore it 👼

apps/web/app/(navigation)/MultiCardLayoutTop.tsx
```diff
@@ -0,0 +1,25 @@
+"use client";
```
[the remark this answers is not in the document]
monika-klokosinska-intive: @PawelPazderski definitely would look better as a single card layout, though if it is too time-consuming to change we can merge as is.

AzulloCyll (approved):

bdeluga (approved):
LGTM ! 🚀
Superb work, to be honest I couldn't have done it the way you did 🤯

zuzannapaw (approved):
Well done! 🚀 Amazing work!

Alessiass (approved):

Alessiass (approved):
Looks great and works as should (I really like graph transitions after you change the period). Approve :rocket:

aniasgit (approved):
Looks and works great 👍

wyola (approved):
Great job! ⭐

monika-klokosinska-intive (approved):

PR #147, opened 2023-06-08
refactor: share budget improvement #P23-386

aniasgit:


apps/web/app/(navigation)/budgets/[id]/BudgetBasinInformation/ShareBudget.tsx
```diff
@@ -90,7 +90,9 @@ export const ShareBudget = ({ budget, onClose }: ShareBudgetProps) => {
   const pageSize = 15;
   const { t, dict } = useTranslate("ShareBudget");
 
-  const initBudgetUsers = budget.budgetUsers?.map((user) => user.id) || [];
+  const budgetUsersWithoutOwner =
+    budget.budgetUsers?.filter((user) => user.id !== budget.userId) || [];
+  const initBudgetUsers = budgetUsersWithoutOwner.map((user) => user.id);
   const [budgetUsers, setBudgetUsers] = useState(initBudgetUsers);
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/BudgetBasinInformation/ShareBudget.tsx:96]: @aniasgit  You can combine the map and filter into a reduce - this way instead of going through the array twice, you will traverse it only once.

```
 const [budgetUsers, setBudgetUsers] = useState(budget.budgetUsers?.reduce((acc: string[], curr) => {
    if (budget.userId !== curr.id) {
      return [...acc, curr.id];
    }
    return acc;
  }, []) || []);
```
Even better: if you dump the reduce inside the state, the function will be run only once, when the state is initialised and not on every re-render.

wyola (approved):
good job!⭐

monika-klokosinska-intive (approved):

PR #150, opened 2023-06-08
refactor: sorting transactions by column - BE integration #P23-403

aniasgit:
I added a mapping of our columns from the transaction table to the column names BE requires in SortDescriptor. This allowed to combine @mpiwonski previous work with the current version of the BE endpoint.


monika-klokosinska-intive (approved):

zuzannapaw (approved):
🚀

PR #151, opened 2023-06-09
feat: export incomes/expenses by email #P23-363

AzulloCyll:
draft

### DIFF
--- a/apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx
+++ b/apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx
@@ -25,6 +25,7 @@ import { ExportResponseProps } from "lib/types";
 import fixCurrencyObject from "lib/validations/fixCurrenyObject";
 import { LinkStyled } from "ui/SideNavigationBar/SubMenu/SubMenu.styled";
 import { ButtonWithDropdown, Icon, Separator, Button } from "ui";
+import { ThemeType } from "ui/theme";
 
 const BudgetContentWrapperStyled = styled.div`
   display: flex;
@@ -56,6 +57,17 @@ const ImportButton = styled(Button)`
   }
   ${device.desktop} {
     padding: 12px 34px;
+    text-decoration: none;
+  }
+`;
+
+const ButtonStyled = styled(Button)`
+  padding: 0;
+  font-weight: normal;
+  text-decoration: none;
+  color: ${({ theme }) => theme.button.primary.main};
+  &:hover {
+    text-decoration: none;
   }
 `;
 
@@ -122,7 +134,7 @@ export const BudgetContent = ({ id }: BudgetsContentProps) => {
             "Content-Type": "application/json",
             Authorization: "Bearer " + session?.user.accessToken,
           },
-          // body: JSON.stringify({ budgetId: id }),
+          body: JSON.stringify({ budgetId: id }),
         }
       )
         .then((res) => console.log(res))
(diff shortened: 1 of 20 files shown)


apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx
```diff
@@ -109,18 +121,40 @@ export const BudgetContent = ({ id }: BudgetsContentProps) => {
...
+            accept: "*/*",
+            "Content-Type": "application/json",
+            Authorization: "Bearer " + session?.user.accessToken,
+          },
+          body: JSON.stringify({ budgetId: id }),
+        }
+      )
+        .then((res) => console.log(res))
+        .catch((err) => console.error(err)),
+    enabled: !!session,
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx:142]: @AzulloCyll change this to `enabled: false` so the query doesn't auto-run

apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx
```diff
@@ -109,18 +121,40 @@ export const BudgetContent = ({ id }: BudgetsContentProps) => {
...
       <Icon icon="file_download" size={12} />
       <span>{tExport(dictExport.exportButtonText)}</span>
     </LinkStyled>
   );
 
-  const emailLink = (
-    <LinkStyled title="email">
+  //trigger
+  const emailButton = (
+    <ButtonStyled onClick={() => exportByMail} title="email" variant="simple">
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx:154]: @AzulloCyll please change this to `() => exportByMail.refetch()}` so the query runs on click.

apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx
```diff
@@ -109,18 +121,40 @@ export const BudgetContent = ({ id }: BudgetsContentProps) => {
     enabled: !!session,
   });
 
+  //query
+  const exportByMail = useQuery({
+    queryKey: ["exporIncomesExpensesByEmail"],
+    queryFn: async () =>
+      await fetch(
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx:128]: @AzulloCyll please use superfetch!

apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx
Comment (AzulloCyll) [reply apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx]: done

apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx
Comment (AzulloCyll) [reply apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx]: done

apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx
Comment (AzulloCyll) [reply apps/web/app/(navigation)/budgets/[id]/BudgetContent/index.tsx]: Done, but I think that error catching and messages in superFetch need a little adjustment. For example: when 404 occurs, message says: "Budget with given id does not exist." I think it need to be changet to somethin more general. 404 means not that we can't found budget, but cant find any resource.

Next thing -> there is no response for status 200 :/

I've showed this error in Toast, but I considering to delete that functionality, because it will be shown when any 404 occurs,

aniasgit (approved):
It works 🚀

monika-klokosinska-intive (approved):

PR #155, opened 2023-06-09
feat: custom categories  #P23-360

bdeluga:
TODO: add nice description


packages/ui/SideNavigationBar/index.tsx
```diff
@@ -37,7 +37,6 @@ const SideNavigationBarStyled = styled.ul<SubMenuBoolean>`
   flex-direction: column;
   align-items: center;
   position: fixed;
-  z-index: 5;
```
Noshi96, line 40: We need this z-index here

image

apps/web/components/CategoryFilter/index.tsx
```diff
@@ -29,36 +32,59 @@ export const CategoryFilter = () => {
...
+                  <CategoryIconStyled small category={category} />
+                  <CategoryTitleStyled>{name}</CategoryTitleStyled>
+                </CheckboxLabelContentStyled>
+              </Checkbox>
+            </li>
+          );
+        })}
+      </CheckboxListStyled>
+      <Separator />
+      <CheckboxListStyled>
```
Noshi96, line 58: On the desktop, the separator has a width of 0. This can be implemented
```tsx
const SeparatorStyled = styled(Separator)`
  width: 100%;
`;
```

apps/web/app/(navigation)/budgets/[id]/TransactionTableController/index.tsx
```diff
@@ -84,9 +88,20 @@ const TransactionTableController = ({ budget }: { budget: BudgetFixed }) => {
         id: item.transactionId.value,
         date: Date.parse(item.budgetTransactionDate),
         amount: item.value,
-        category: categoryMap[item.categoryType]
-          ? categoryMap[item.categoryType]
-          : categoryMap.HomeSpendings,
+        //try finding custom category
+        category: userCategories.find(
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/TransactionTableController/index.tsx:92]: @bdeluga woah, this can be simplified with an alternative
```
   category: userCategories.find(
              (category) => category.name! === item.categoryType.categoryName
            ) || categoryMap[item.categoryType.categoryName] || categoryMap.HomeSpendings,
```

`find` returns undefined if nothing was found, so alternative will start evaluating the next element.

apps/web/app/(navigation)/budgets/[id]/CreateNewTransaction/index.tsx
```diff
@@ -226,17 +227,42 @@ export const CreateNewTransaction = ({
...
-                            <>
-                              <CategoryIcon small category={category} />
-                              <span>{category.name}</span>
-                            </>
-                          ),
-                        })
-                      )}
+                      items={
+                        userCategories.length
+                          ? [
```
Comment (monika-klokosinska-intive) [inline apps/web/app/(navigation)/budgets/[id]/CreateNewTransaction/index.tsx:232]: @bdeluga you should be fine without this check.
```
[...userCategories.map((category) => ({
                                value: category.name!,
                                label: (
                                  <>
                                    <CategoryIcon small category={category} />
                                    <span>{category.name}</span>
                                  </>
                                ),
                              })),
                              ...Object.entries(categoryMap).map(
                                ([categoryKey, category]) => ({
                                  value: categoryKey,
                                  label: (
                                    <>
                                      <CategoryIcon small category={category} />
                                      <span>{category.name}</span>
                                    </>
                                  ),
                                })
                              ),
                            ]
```
Assuming `userCategories` is an empty array in the worst case scenario (and you already assumed that since you are checking `length` and running `map` on it ;)), you can still spread it. It won't add anything to the final array.

zuzannapaw (commented):
I spotted a problem when after adding new category, I want to add a transaction with that category and when the transaction appears on table, you can only see the default icon (house) and colors. Only after refreshing list ,the proper icon appears. Besides that, it works great, well done for your hard work 🚀 👏 After adding @monika-klokosinska-intive  's suggestions and fixing this problem , I think it would be ready to merge.

apps/web/app/(navigation)/budgets/[id]/TransactionTableController/index.tsx
Comment (bdeluga) [reply apps/web/app/(navigation)/budgets/[id]/TransactionTableController/index.tsx:92]: Thank god there is a simplification 🚀

apps/web/app/(navigation)/budgets/[id]/CreateNewTransaction/index.tsx
Comment (bdeluga) [reply apps/web/app/(navigation)/budgets/[id]/CreateNewTransaction/index.tsx:232]: Thanks for simplifying this 🚀Working under such pressure leads to a lot of spaghetti code 🍝

apps/web/components/CategoryFilter/index.tsx
bdeluga: Alright ! 🚀

packages/ui/SideNavigationBar/index.tsx
bdeluga: So that's what it was for... Should've checked after deleting... 🙃

bdeluga:
> I spotted a problem when after adding new category, I want to add a transaction with that category and when the transaction appears on table, you can only see the default icon (house) and colors. Only after refreshing list ,the proper icon appears. Besides that, it works great, well done for your hard work 🚀 👏 I also added french. After adding @monika-klokosinska-intive 's suggestions and fixing this problem , I think it would be ready to merge.

I surely has to try figuring something about it today. Thanks for translation Madame 🐸

Noshi96 (approved):
Awesome 🤩

zuzannapaw (approved):
Well done with fix, now it works like a charm, congrats 🚀

PR #157, opened 2023-06-09
feat: export budgets via mail #P23-361

Noshi96:
Ticket: https://tracker.intive.com/jira/browse/PATRO23-361

By selecting send by email from the export dropdown option, we can export our budgets. After a successful operation, the budgets as a CSV file will be sent to our email address, and we will see a success toast.


monika-klokosinska-intive (approved):

zuzannapaw (approved):
looks and works good! 🚀

